PR #103, opened 2023-01-12
NH-29850 unify OTel service name with SW_APM service key

tammy-baylis-swi:
This implements `"Idea #2"`, the unification of `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, and `SW_APM_SERVICE_KEY` for setting `service.name` in OTel resource attributes used by the configured TracerProvider. Details are here: https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3379855366/OTel+vs+SWO+Service+Name?focusedCommentId=3388899356#Idea-#2:-Unify-OTEL_SERVICE_NAME-and-SW_APM_SERVICE_KEY

Summary: `SolarWindsApmConfig` is responsible for `_calculate_service_name` and uses a fresh OTel `Resource.create` (by default) to get OTel's calculation of `service.name` based on the `OTEL_*` env vars if provided. If there isn't an OTel service name, we use `SW_APM_SERVICE_KEY` instead. ApmConfig also updates its `service_key` value using that name. Then the Configurator uses ApmConfig to (1) init liboboe reporter with the updated `service_key`, and (2) `Resource.create` again with the updated `service_name` and send that one to TracerProvider.

I've done some manual testing and documented results here: https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3388801105/2023-01-12+Unification+of+OTEL+and+SW+APM+service.name More unit tests added too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/solarwindscloud/solarwinds-apm-python/compare/rel-0.4.0...HEAD)
 ### Added
 - Add support for OTEL_SERVICE_NAME, OTEL_RESOURCE_ATTRIBUTES environment variables ([#90](https://github.com/solarwindscloud/solarwinds-apm-python/pull/90))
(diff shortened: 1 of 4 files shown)


cheempz (approved):
LGTM, really appreciate the extended regression suite _and_ manual tests @tammy-baylis-swi!

PR #106, opened 2023-01-25
NH-29063 PoC: set_transaction_name API

tammy-baylis-swi:
Proof-of-concept/conversation starter for how we might try to set up a `set_transaction_name` method for customers to use. Run using this testbed update: https://github.com/appoptics/solarwinds-apm-python-testbed/pull/47

**Summary of changes:**
1. Adds new `SolarWindsTxnNameCustomizer`, which for now is separate and used slightly differently than the existing `SolarWindsTxnNameManager`
2. `SolarWindsTxnNameCustomizer` is initialized by Configurator and linked to the Inbound Metrics Processor
3. Inbound Metrics Processor uses TxnNameCustomizer to retrieve any cached custom txn names by `trace_id` (versus by `trace_id` and `span_id` when setting in TxnNameManager)
4. Inbound Metrics Processor prioritizes custom txn name over http/regular span name for liboboe call and for caching for exporter to use later.
5. TxnNameCustomizer is updated by new API call `set_transaction_name`, by getting the current TracerProvider at execution time and finding the Inbound Metrics Processor.

**Single example trace:**
* under Django A transaction "my-foo-name": https://my.na-01.st-ssp.solarwinds.com/136444677275740160/traces/5FEB13C065A5F33BAA17608F140478B7/3F1C9D6B963B0182/details
* under Django B transaction "my-bar-name": https://my.na-01.st-ssp.solarwinds.com/136444677275740160/traces/5FEB13C065A5F33BAA17608F140478B7/DD9AB8C89BB3E79C/details

**Thoughts**
I don't think we can use this approach alone to e.g. call `set_transaction_name` in Django B then get the root server span of Django A to have a customize txn name in distributed trace. (Example trace where Django B calls API but Django A does not: [at Django B service](https://my.na-01.st-ssp.solarwinds.com/136444677275740160/traces/6D8D9F2122A33C171F78DAE5D0193E4F/5F00494FC1F20D62/details), [at Django A service](https://my.na-01.st-ssp.solarwinds.com/136444677275740160/traces/6D8D9F2122A33C171F78DAE5D0193E4F/9759D277BF8E48E4/details).)

This makes sense since each service will have its own inbound metrics processor and exporter running. If we wanted them to talk to each other, instrumented services would have to read response headers(???)

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -1,0 +1,38 @@
+# © 2023 SolarWinds Worldwide, LLC. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at:http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
+"""
+Example usage:
+
+from solarwinds_apm.api import set_transaction_name
+set_transaction_name("my-foo-name")
+"""
+
+import logging
+
+from opentelemetry.trace import get_current_span, get_tracer_provider
+
+from solarwinds_apm.inbound_metrics_processor import SolarWindsInboundMetricsSpanProcessor
+
+logger = logging.getLogger(__name__)
+
+def set_transaction_name(name: str) -> None:
+    # Assumes TracerProvider's active span processor is SynchronousMultiSpanProcessor
+    # or ConcurrentMultiSpanProcessor
+    span_processors = get_tracer_provider()._active_span_processor._span_processors
+    inbound_processor = None
+    for spr in span_processors:
+        if type(spr) == SolarWindsInboundMetricsSpanProcessor:
+            inbound_processor = spr
+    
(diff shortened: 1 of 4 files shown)


cheempz:
This looks great @tammy-baylis-swi!  I agree this is how we should support custom transaction naming first.  A few thoughts on the work to make it real:

* Your PR description has made me realize that the need to cache txn name per trace ID + span ID (rather than just trace ID) is not necessary in the typical case where each service in a distributed trace will be a separate instrumented process, each with it's own span processor / exporter etc.  So in practice a span processor and exporter would _only_ see the spans for its own service (and not any from a remote upstream).  The special case I can think of is the "reentrant" or "recursive" call, where an application is set up so that one of its endpoints ends up calling another, i.e. `Service A /endpoint1 -- RPC call --> Service A /endpoint2`.  In this case it's the same process handling the "distributed" workflow.  It seems strange but we've actually run into customer setup in the past like this so we probably do still need to account for it.

* We'd want to support the `set_transaction_name` API call being made multiple times within a process, the last call made should win.  I think your POC already allows that, so it may just be a matter of debug logging and any optimizing that might make sense.

cheempz:
One more thought while looking at documentation updates: there's already a non-OTel API implemented by our distro, the `solarwinds_apm.apm_ready` call... may want to consolidate into the `api` module in this PoC.

tammy-baylis-swi:
Closing this PoC as replaced with https://github.com/solarwindscloud/solarwinds-apm-python/pull/115

PR #115, opened 2023-02-27
NH-32028 API method to set custom txn name

tammy-baylis-swi:
This adds a new API method, `set_transaction_name`, to Python APM library so that customers can assign a custom txn name to a trace from their code.

Also moves existing `solarwinds_ready` to the same, new `api` module. The old `solarwinds_ready` is now no-op and with deprecation warnings.

`set_transaction_name` should work like what's described for .NET library ([this customer doc](https://documentation.solarwinds.com/en/success_center/observability/content/configure/services/dotnet/sdk.htm#trace-settransactionname)). Detailed description and results of some manual testing here: https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3468230856/2023-02-23+Custom+Transaction+Naming+API, using [these testbed updates](https://github.com/appoptics/solarwinds-apm-python-testbed/pull/47).

The way this works with OTel is:
1. Inbound metrics span processor has a new `on_start` method override that always caches the trace_id and span_id of the "current" service entry span into mutable OTel [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/), associated with the much less mutable OTel context.
2. Yes, the way I'm writing to baggage is a little questionable. Waiting for an answer [here on CN Slack](https://cloud-native.slack.com/archives/C01PD4HUVBL/p1677181444735639).
3. If the customer's code calls our API `set_transaction_name` method, we use OTel tracer to access the inbound metrics span processor and the cached current service entry span from baggage. Customer custom txn name is written to the former's transaction name manager with the entry span from the latter as key.
4. Inbound metrics span processor `on_end` now checks for cached custom txn name for each span, and uses them with priority over the existing http or 'regular' txn name if exists for the current ended span.
5. The last span to end with an associated call to `set_transaction_name` is the last custom name assigned by the customer, and will be the one used over any earlier calls to the API in older spans.

Polished customer documentation will be completed via https://swicloud.atlassian.net/browse/NH-34187. Also this completely replaces this early [PoC](https://github.com/solarwindscloud/solarwinds-apm-python/pull/106).

Please let me know if any questions or suggestions  😃


tammy-baylis-swi:
Ready now!

solarwinds_apm/inbound_metrics_processor.py
```diff
@@ -141,12 +170,25 @@ def calculate_transaction_names(
...
             trans_name = span.name
         return trans_name, url_tran
 
+    def calculate_custom_transaction_name(self, span: "ReadableSpan") -> Any:
+        """Get custom transaction name for trace by trace_id, if any"""
+        trans_name = None
+        trace_span_id = f"{span.context.trace_id}-{span.context.span_id}"
+        custom_name = self.apm_txname_manager.get(trace_span_id)
+        if custom_name:
+            trans_name = self.apm_txname_manager[trace_span_id]
```
raphael-theriault-swi, line 189: i'm not sure since not too familiar with python but couldn't this be
```suggestion
            trans_name = custom_name
```

raphael-theriault-swi (commented):
Looks good, just had that one small question

solarwinds_apm/inbound_metrics_processor.py
tammy-baylis-swi: 🙃 Yup! Addressed in [160180d](https://github.com/solarwindscloud/solarwinds-apm-python/pull/115/commits/160180d5ba6da0cf8638a6a642515864553ce23d) 😺

raphael-theriault-swi (approved):

PR #119, opened 2023-03-07
NH-35046 No longer need PLATFORM env var for sdist installation, tests

tammy-baylis-swi:
Fixes an issue where `PLATFORM` environment variable is required by customers(!) to install Python APM library from sdist, i.e. on Alpine Linux. Thank you @cheempz for spotting this.

`PLATFORM` is no longer mandatory for extension creation by setuptools (setup.py) nor testing. I published this as [TestPyPI version 0.8.1.4](https://test.pypi.org/project/solarwinds-apm/0.8.1.4/) and I tested installation of this with:

1. Change [this line](https://github.com/appoptics/solarwinds-apm-python-testbed/blob/main/flask_testpypi_app/Dockerfile#L9) in testbed to `RUN pip install --extra-index-url https://test.pypi.org/simple/ --no-binary solarwinds-apm solarwinds-apm==0.8.1.4` to force installation from sdist instead of wheels.
2. Build test app without setting `ENV` for platform: `docker-compose -f docker-compose.flask-from-testpypi.yml build --no-cache`
3. No longer gets pip error at build

Install tests pass after removing `PLATFORM` from [verify_install.yaml](https://github.com/solarwindscloud/solarwinds-apm-python/pull/119/files#diff-cac234b3030c63d11179b2ea77aa921bb112b34480a4313e874bd79f3848965d): https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/4359203186

I can run the testbed without `PLATFORM` now; changes in [this PR](https://github.com/appoptics/solarwinds-apm-python-testbed/pull/52).

`PLATFORM` is optional for some `make` actions (Makefile): built wheel tagging (in GH workflows for releases or local `make package`) and lambda builds. If not provided, `PLATFORM` defaults to `x86_64`. `PLATFORM=aarch64` is provided by the `build_publish_aarch64` GH actions that build aarch64 on QEMU (where `uname -m` doesn't seem to be `aarch64`).

This also makes the `--no-binary` / `--only-binary` pip calls in the install tests less ambiguous/more correct.

### DIFF
--- a/tests/docker/install/_helper_check_wheel.sh
+++ b/tests/docker/install/_helper_check_wheel.sh
@@ -91,7 +91,7 @@ function get_wheel(){
             exit 1
         fi
     else
-        pip_options=(--only-binary solarwinds-apm --dest "$wheel_dir")
+        pip_options=(--only-binary solarwinds-apm --dest "$wheel_dir" solarwinds-apm)
         if [ "$MODE" == "testpypi" ]
         then
             pip_options+=(--extra-index-url https://test.pypi.org/simple/)
--- a/tests/docker/install/_helper_check_wheel.sh
+++ b/tests/docker/install/_helper_check_wheel.sh
@@ -40,18 +40,11 @@ VALID_PLATFORMS=(
     "x86_64"
     "aarch64"
 )
-if [ -z "$PLATFORM" ]
-then
-  echo "WARNING: Did not provide PLATFORM for check_wheel test."
-  echo "Defaulting to PLATFORM=x86_64"
-  PLATFORM=x86_64
-fi
+PLATFORM=$(uname -m)
 if [[ ! " ${VALID_PLATFORMS[*]} " =~ ${PLATFORM} ]]
 then
-  echo "FAILED: Did not provide valid PLATFORM for check_wheel test. Must be one of: x86_64, aarch64."
+  echo "FAILED: Invalid platform for check_wheel test. Must be run on one of: x86_64, aarch64."
   exit 1
-else
-  echo "Using provided PLATFORM=$PLATFORM for check_wheel test."
 fi
 
 if [ "$PLATFORM" == "x86_64" ]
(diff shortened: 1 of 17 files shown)


tests/docker/install/_helper_check_wheel.sh
```diff
@@ -101,7 +91,7 @@ function get_wheel(){
             exit 1
         fi
     else
-        pip_options=(--only-binary solarwinds-apm --dest "$wheel_dir")
+        pip_options=(--only-binary solarwinds-apm --dest "$wheel_dir" solarwinds-apm)
```
cheempz, line 94: Minor is the addition of `solarwinds-apm` needed? Seeing later that we do seem to append it to pip_options.

cheempz (commented):
LGTM pending minor question, thanks for addressing this so quickly @tammy-baylis-swi!

tests/docker/install/_helper_check_wheel.sh
tammy-baylis-swi: I think it is, technically. pip `--only-binary` and `--no-binary` are supposed to have a `<format_control>` value provided: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-only-binary

The way it's currently written in the install tests still works and I'm not sure why given the above. The install tests pass on GH (have been for a long time) and this also passes on my local on `main` branch: `MODE=testpypi docker-compose run --rm py3.7-install-alpine3.12`.

The missing `<format_control>` becomes an issue when I try to install from the testbed for `flask-from-testpypi`. Both of these error:

```
pip install --no-binary solarwinds-apm==0.8.0
pip install --no-binary solarwinds-apm
```
But this is ok:
```
pip install --no-binary solarwinds-apm==0.8.0 solarwinds-apm
```
Actually this is failing too...
```
pip install --no-binary solarwinds-apm solarwinds-apm
```

...Ok. I'm going to revert [b7fcfd6](https://github.com/solarwindscloud/solarwinds-apm-python/pull/119/commits/b7fcfd66c1874c41f2436069975b5dfed7ef57ce) and do a fresh publish and test.

tammy-baylis-swi:
Thanks for first review @cheempz ! I've reverted [b7fcfd6](https://github.com/solarwindscloud/solarwinds-apm-python/pull/119/commits/b7fcfd66c1874c41f2436069975b5dfed7ef57ce), published to TestPyPI as 0.8.1.5, and run the install tests again which seem ok: https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/4369183330

Please could you have another look and let me know if anything else needed 😃

cheempz (approved):
LGTM, thanks for the revisit @tammy-baylis-swi!

PR #126, opened 2023-03-20
NH-36210 Bugfix set_transaction_name when agent_enabled False + linting

tammy-baylis-swi:
Fixes errors at customer code calls to API `set_transaction_name` when `agent_enabled` False!:

```
  File "/app/main.py", line 79, in reentry_async
    set_transaction_name("custom-name-that-shouldnt-be-used")
  File "/solarwinds-apm-python/solarwinds_apm/api/__init__.py", line 48, in set_transaction_name
    get_tracer_provider()._active_span_processor._span_processors
AttributeError: 'NoOpTracerProvider' object has no attribute '_active_span_processor'
```

No longer errors when e.g. `Incorrect service key format.`, testbed ASGI app is run, and `curl -v http://0.0.0.0:8005/reentry_async` which includes calls to `set_transaction_name`.

Trace still exports when agent enabled and `set_transaction_name` calls work on entry spans: https://my.na-01.cloud.solarwinds.com/140638900734749696/traces/0E45BB4C6DD24EB273C1323B6C28D9EF/33278EA380018F84/details/breakdown/33278EA380018F84/rawData

Also fixes several linting errors that started happening with the liboboe imports (updated to latest tox dependency virtualenv `'20.20.0'->'20.21.0'`) by ignoring them.

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -8,7 +8,7 @@ import logging
 from typing import Any
 
 from opentelemetry import baggage
-from opentelemetry.trace import get_tracer_provider
+from opentelemetry.trace import get_tracer_provider, NoOpTracerProvider
 
 from solarwinds_apm.apm_constants import (
     INTL_SWO_CURRENT_SPAN_ID,
@@ -41,6 +41,13 @@ def set_transaction_name(custom_name: str) -> bool:
      from solarwinds_apm.api import set_transaction_name
      result = set_transaction_name("my-foo-name")
     """
+    if isinstance(get_tracer_provider(), NoOpTracerProvider):
+        logger.debug(
+            "Cannot cache custom transaction name %s because agent not enabled; ignoring",
+            custom_name,
+        )
+        return False
+
     # Assumes TracerProvider's active span processor is SynchronousMultiSpanProcessor
     # or ConcurrentMultiSpanProcessor
     span_processors = (
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -8,13 +8,15 @@ import logging
 from typing import Any
 
 from opentelemetry import baggage
-from opentelemetry.trace import get_tracer_provider, NoOpTracerProvider
+from opentelemetry.trace import NoOpTracerProvider, get_tracer_provider
 
 from solarwinds_apm.apm_constants import (
     INTL_SWO_CURRENT_SPAN_ID,
     INTL_SWO_CURRENT_TRACE_ID,
 )
 from solarwinds_apm.apm_oboe_codes import OboeReadyCode
+
+# pylint: disable=import-error,no-name-in-module
 from solarwinds_apm.extension.oboe import Context
 from solarwinds_apm.inbound_metrics_processor import (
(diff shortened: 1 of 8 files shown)


solarwinds_apm/api/__init__.py
```diff
@@ -41,6 +43,13 @@ def set_transaction_name(custom_name: str) -> bool:
      from solarwinds_apm.api import set_transaction_name
      result = set_transaction_name("my-foo-name")
     """
+    if isinstance(get_tracer_provider(), NoOpTracerProvider):
+        logger.debug(
+            "Cannot cache custom transaction name %s because agent not enabled; ignoring",
+            custom_name,
+        )
+        return False
```
molafson, line 51: I don't think we have documented recommendations for return values for agents in no-op mode. For the AppOptics java agent and the .net library when in no-op methods return `True` except for methods like `IsAgentAvailable`. 

@cheempz should there be a standardized behavior for no-op methods?

cheempz: Good point, I guess the rationale for AO Java/.NET returning `True` in almost all cases is to minimize the chance that in no-op mode our API calls might cause unnecessary handling of error response in the customer's application code making these calls.

It looks like SWO Java is also returning False when in no-op mode https://github.com/appoptics/solarwinds-apm-java/blob/main/solarwinds-otel-sdk/src/main/java/com/appoptics/api/ext/SolarWindsAgent.java#L37 so we have mini-standardization between the custom distros ;) But given these are our own API we can get them to all agree -- I do lean towards returning True across languages in this case. WDYT @tammy-baylis-swi @CleverChuk?

tammy-baylis-swi: > minimize the chance that in no-op mode our API might cause unnecessary handling of error response in the customer's application code making these calls

This makes sense to me! 👍 `set_transaction_name` now returns True for any case in [3d503c9](https://github.com/solarwindscloud/solarwinds-apm-python/pull/126/commits/3d503c90973efd0df7869341f47cc5c71711de3b)

tammy-baylis-swi: Correction in [fbf687b](https://github.com/solarwindscloud/solarwinds-apm-python/pull/126/commits/fbf687bb55032977e1fba2bc68542dd6c2ad3b45)!:

[Agent not enabled](https://github.com/solarwindscloud/solarwinds-apm-python/blob/fbf687bb55032977e1fba2bc68542dd6c2ad3b45/solarwinds_apm/api/__init__.py#L51) --> True
[Processor not configured](https://github.com/solarwindscloud/solarwinds-apm-python/blob/fbf687bb55032977e1fba2bc68542dd6c2ad3b45/solarwinds_apm/api/__init__.py#L66) --> False
[Span not started](https://github.com/solarwindscloud/solarwinds-apm-python/blob/fbf687bb55032977e1fba2bc68542dd6c2ad3b45/solarwinds_apm/api/__init__.py#L75) --> False
[Able to cache custom name](https://github.com/solarwindscloud/solarwinds-apm-python/blob/fbf687bb55032977e1fba2bc68542dd6c2ad3b45/solarwinds_apm/api/__init__.py#L83) --> True

molafson (approved):
Looks good 👍

PR #136, opened 2023-04-21
NH-34752 Parse txn filters and calculate tracing mode

tammy-baylis-swi:
This is more steps towards implementing transaction filtering:

1. Update `update_transaction_filters` in the Configurator to use existing helper `OboeTracingMode.get_oboe_trace_mode` to map filter's `tracing` to int for liboboe
2. Also more checks for valid filter's `regex` values
3. Adds `calculate_tracing_mode` used by sampler to apply per-request filter or "global" `tracing_mode` that can also be configured.

This only filters for span `<kind>:<name>` right now because `http.*` attributes generated by instrumentation libraries are not yet available to the sampler, but this is gradually being implemented in [OTel #936](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/936).

Example basic config file JSON I'm using with Django A:
```
{
    "tracingMode": "enabled",
    "transactionSettings": [
        {
            "regex": "SERVER:another/",
            "tracing": "disabled"
        },
    ]
}
```

Next PRs:
* Add two-signal check ~~OR support for third filter component, signal (see Slack thread)~~ See https://github.com/solarwindscloud/solarwinds-apm-python/pull/137
* Add more unit tests
* Add web request filtering -- best tested when some [OTel #936](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/936) updates are merged and released
* infodev/customer docs update

Please let me know if any questions/suggestions!

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -569,15 +569,16 @@ class SolarWindsApmConfig:
                     )
                     continue
 
+                txn_filter_re = None
                 try:
-                    re.compile(filter["regex"])
+                    txn_filter_re = re.compile(filter["regex"])
                 except re.error:
                     logger.warning(
                         "Transaction filter regex invalid. Ignoring: %s",
                         filter,
                     )
                     continue
-                txn_filter["regex"] = filter["regex"]
+                txn_filter["regex"] = txn_filter_re
                 self.__config["transaction_filters"].append(txn_filter)
 
         logger.debug(
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -7,6 +7,7 @@
 import json
 import logging
 import os
+import re
 import sys
 from collections import defaultdict
 from functools import reduce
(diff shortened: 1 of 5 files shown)


solarwinds_apm/apm_config.py
```diff
@@ -549,7 +550,35 @@ def update_transaction_filters(self, cnf_dict: dict) -> None:
...
+
+                if not len(filter["regex"]) > 0:
+                    logger.warning(
+                        "Transaction filter regex must not be empty. Ignoring: %s",
+                        filter,
+                    )
+                    continue
+
+                try:
+                    re.compile(filter["regex"])
```
raphael-theriault-swi, line 573: Since you're already compiling the regex in here wouldn't it make sense to store the compiled version in `txn_filter` so it doesn't have to get recompiled on every match

tammy-baylis-swi: Good idea! Changed in [d28c040](https://github.com/solarwindscloud/solarwinds-apm-python/pull/136/commits/d28c040a158f4349bb918285dd92e04fda31cc80).

raphael-theriault-swi (approved):
Looks good !

PR #137, opened 2023-04-25
NH-34752 Add construct_url for calculate_tracing_mode

tammy-baylis-swi:
Adds `construct_url` helper for `calculate_tracing_mode`. I've done it this way so that if there is a URL, it gets used as the identifier and the customer has no choice for filtering (see also [Slack thread](https://swicloud.slack.com/archives/C2J43PE4R/p1682363645105209?thread_ts=1682098813.423939&cid=C2J43PE4R)). At the moment it will always return empty string until a new version of `opentelemetry-python-contrib` is released, so more testing will come later. This is set to merge into https://github.com/solarwindscloud/solarwinds-apm-python/pull/136.

### DIFF
--- a/solarwinds_apm/sampler.py
+++ b/solarwinds_apm/sampler.py
@@ -92,14 +92,19 @@ class _SwSampler(Sampler):
         #   https://github.com/open-telemetry/opentelemetry-python-contrib/issues/936
         if not attributes:
             return ""
+
+        url = ""
         scheme = attributes.get(SpanAttributes.HTTP_SCHEME)
         host = attributes.get(SpanAttributes.NET_HOST_NAME)
+        port = attributes.get(SpanAttributes.NET_HOST_PORT)
         target = attributes.get(SpanAttributes.HTTP_TARGET)
-        if scheme and host and target:
+        if scheme and host and target and port:
+            url = f"{scheme}://{host}:{port}{target}"
+            logger.debug("Constructed url for filtering: %s", url)
+        elif scheme and host and target:
             url = f"{scheme}://{host}{target}"
             logger.debug("Constructed url for filtering: %s", url)
-            return url
-        return ""
+        return url
 
     def calculate_tracing_mode(
         self,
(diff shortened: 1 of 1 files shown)


raphael-theriault-swi (commented):
Does python use `NET_HOST_PORT` ?

tammy-baylis-swi:
> Does python use `NET_HOST_PORT` ?

@raphael-theriault-swi Yes!: https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/trace/__init__.py#L356 And this semconv is used by several instrumentation libraries from a quick grep.

So maybe something like this, instead of without port as currently?

```
url = f"{scheme}://{host}:{port}{target}"
```

solarwinds_apm/sampler.py
```diff
@@ -81,6 +82,26 @@ def __init__(self, apm_config: "SolarWindsApmConfig"):
...
+        #   availability after OTel instrumentation library updates released
+        #   https://github.com/open-telemetry/opentelemetry-python-contrib/issues/936
+        if not attributes:
+            return ""
+        scheme = attributes.get(SpanAttributes.HTTP_SCHEME)
+        host = attributes.get(SpanAttributes.NET_HOST_NAME)
+        port = attributes.get(SpanAttributes.NET_HOST_PORT)
+        target = attributes.get(SpanAttributes.HTTP_TARGET)
+        if scheme and host and target:
+            url = f"{scheme}://{host}:{port}{target}"
```
raphael-theriault-swi, line 100: My understanding is that `NET_HOST_PORT` is only required if it doesn't match the scheme (ie 80 for http and 443 for https) so it should probably only be added to the url if present to not end up with urls like `http://google.com:/search`

tammy-baylis-swi: Updated in [acd71d8](https://github.com/solarwindscloud/solarwinds-apm-python/pull/137/commits/acd71d8f4ace9c048b2f3240e6df2746333b8ebd).

raphael-theriault-swi (approved):
LGTM !

PR #143, opened 2023-04-28
NH-34752 Add sw.span_name attribute at export

tammy-baylis-swi:
Add `sw.span_name` attribute at export as per [acceptance criteria](https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3538485339/Transaction+Filtering+aka+per-request+tracing+mode#Acceptance-Criteria).

I've added the test case's additional expected call `mocker.call("TransactionName", "foo")` because we've gone from a mock txn name manager with unspecified `get` method (which does magic mocking) to an explicit one.

[Example trace](https://my.na-01.cloud.solarwinds.com/140638900734749696/traces/C2B6EB747B271FAE44CF17A17E7F1A6C/5267657C46A48D38/details/breakdown/5267657C46A48D38/raw-data?perspective=requests&serviceId=e-1540126275982954496) with `sw.span_name` included in the exported spans.


cheempz (approved):

PR #154, opened 2023-05-29
NH-41991 Add stub for verify_install_arm64.yaml

tammy-baylis-swi:
Adds stub for verify_install_arm64.yaml that uses https://github.com/solarwindscloud/ec2-runner-action. Adds two jobs in matrix that will `echo` to make sure new secrets work. After this PR to `main` I can run the action with dispatch and update it.

### DIFF
--- a/.github/workflows/verify_install_arm64.yaml
+++ b/.github/workflows/verify_install_arm64.yaml
@@ -71,6 +71,7 @@ jobs:
       - run: echo "Hello from EC2 ${{ matrix.job }} !"
 
   terminate-arm64:
+    if: ${{ always() }}
     needs:
       - launch-arm64
       - install-tests
--- a/.github/workflows/verify_install_arm64.yaml
+++ b/.github/workflows/verify_install_arm64.yaml
@@ -62,9 +62,10 @@ jobs:
     needs:
       - launch-arm64
     strategy:
-      matrix: |
-        py3.7-ubuntu18.04
-        py3.11-ubuntu20.04
+      matrix:
+        job:
+          - py3.7-ubuntu18.04
+          - py3.11-ubuntu20.04
     runs-on: ${{ fromJSON(needs.launch-runners.outputs.matrix)[matrix.job].label }}
     steps:
       - run: echo "Hello from EC2 ${{ matrix.job }} !"
--- a/.github/workflows/verify_install_arm64.yaml
+++ b/.github/workflows/verify_install_arm64.yaml
@@ -37,7 +37,7 @@ jobs:
     outputs:
       matrix: ${{ steps.launch.outputs.matrix }}
     steps:
-      - uses: aws-actions/configure-aws-credentials@v1
+      - uses: aws-actions/configure-aws-credentials@v2
         with:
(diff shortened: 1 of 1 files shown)


cheempz (dismissed):

.github/workflows/verify_install_arm64.yaml
```diff
@@ -0,0 +1,87 @@
...
+          subnet-id: subnet-0fd499f8a50e41807
+          security-group-ids: sg-0fd8d8cd6effda4a5
+
+  install-tests:
+    needs:
+      - launch-arm64
+    strategy:
+      matrix: |
+        py3.7-ubuntu18.04
+        py3.11-ubuntu20.04
```
raphael-theriault-swi, line 67: The matrix syntax for the EC2 action and [the one for actual jobs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) are different since you can't have structured values as action inputs. This would have to look something like this to work, there's also a [VSCode extension for workflow syntax](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions) that's really helpful for catching a lot of stuff in advance that is impossible to check locally.
```suggestion
      matrix:
        job:
          - py3.7-ubuntu18.04
          - py3.11-ubuntu20.04
```

raphael-theriault-swi (changes requested):

.github/workflows/verify_install_arm64.yaml
tammy-baylis-swi: Thanks @raphael-theriault-swi !! Changed in [b2575b2](https://github.com/solarwindscloud/solarwinds-apm-python/pull/154/commits/b2575b28c5cb18928681c25c38ea718435f9f8d4), and I've got that helpful extension installed now.

.github/workflows/verify_install_arm64.yaml
```diff
@@ -0,0 +1,88 @@
...
+    strategy:
+      matrix:
+        job:
+          - py3.7-ubuntu18.04
+          - py3.11-ubuntu20.04
+    runs-on: ${{ fromJSON(needs.launch-runners.outputs.matrix)[matrix.job].label }}
+    steps:
+      - run: echo "Hello from EC2 ${{ matrix.job }} !"
+
+  terminate-arm64:
```
raphael-theriault-swi, line 73: Woops missed that one, this should always run even if the tests fail
```suggestion
  terminate-arm64:
    if: always()
```

raphael-theriault-swi (changes requested):

.github/workflows/verify_install_arm64.yaml
tammy-baylis-swi: I added it with brackets in [5195485](https://github.com/solarwindscloud/solarwinds-apm-python/pull/154/commits/51954854fd5e26a6eba311f50a4dc8bf9061f6e5) like the [repo readme](https://github.com/solarwindscloud/ec2-runner-action).

raphael-theriault-swi (approved):
Thanks for the changes !

PR #156, opened 2023-05-31
NH-41991 Verify Installation tests use ec2 action runners

tammy-baylis-swi:
Replaces old `verify_installation` workflow that used regular GH action runners + QEMU with new `verify_installation` that uses our [ec2-runner-action](https://github.com/solarwindscloud/ec2-runner-action). Thank you Raph for all your tips through this one!

Much faster and no more random failures! Example run: https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/5135753921

Both x86 and aarch64 wheel installations are being tested, e.g.:

```
# install-tests (amazonlinux:2023, x64)

Using provided MODE=pypi for check_wheel test.
Set WHEEL_FILENAME=manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl based on platform for check_wheel test.
```

```
# install-tests (amazonlinux:2023, arm64)

Using provided MODE=pypi for check_wheel test.
Set WHEEL_FILENAME=manylinux_2_27_aarch64.manylinux_2_28_aarch64.whl based on platform for check_wheel test.
```

The Python version X Linux distro coverage is similar to the old workflow. Neither workflow has much py3.10/3.11-on-Ubuntu coverage because of https://swicloud.atlassian.net/browse/NH-31430. py3.12 isn't covered at all because it hasn't been released yet, though an alpha version is available in some public images. I will address these in future PRs.


raphael-theriault-swi (approved):
Running a tiny bit faster I see

PR #158, opened 2023-05-31
NH-31430 Add Ubuntu 22, use current setuptools in install tests

tammy-baylis-swi:
Removes the part of the install_tests script that uninstalls the latest `setuptools` and installs an older version, because this is no longer needed. `opentelemetry-bootstrap` now works ok with current `setuptools`.

Also adds local and GH install test containers for Ubuntu 22 and py3.10/3.11. Ubuntu 22 does not need to use `deadsnakes` to install Python but Ubuntu 20 still needs it.

GH test run: https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/5138153171

### DIFF
--- a/.github/workflows/verify_install.yaml
+++ b/.github/workflows/verify_install.yaml
@@ -52,6 +52,7 @@ jobs:
             registry.fedoraproject.org/f33/python3
             ubuntu:18.04
             ubuntu:20.04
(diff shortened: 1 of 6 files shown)


jerrytfleung (approved):
LGTM

PR #160, opened 2023-06-01
NH-41991 Fix install test Python version coverage

tammy-baylis-swi:
Fixes install test Python version coverage by 'rotating' the workflow matrix.

Before this change ([test run](https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/5138153171)), as an example problem there should have been 4 `ubuntu:18.04` runs for
1. py3.7 x86
2. py3.7 arm64
3. py3.8 x86
4. py3.8 arm64

but only 2 ran for py3.8 on each x86, arm64. 36 jobs ran (2 setup + 34 test runs).

After this change ([test run](https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/5149013274)), we now get all four `ubuntu:18.04` runs and 46 jobs. This is by changing `strategy > matrix > image` to the full list of `strategy > matrix > hostname`, and changing `strategy > matrix > include` to specify the images used.

Please let me know what you think 🙂

### DIFF
--- a/.github/workflows/verify_install.yaml
+++ b/.github/workflows/verify_install.yaml
@@ -87,32 +87,38 @@ jobs:
     runs-on: ${{ matrix.arch == 'arm64' && fromJSON(needs.launch-arm64.outputs.matrix)[matrix.image].label || 'ubuntu-latest' }}
     strategy:
       matrix:
-        image:
-          - amazonlinux:2023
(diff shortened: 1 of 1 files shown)


CleverChuk (approved):
I have no clue what's going on, however LGTM!

PR #162, opened 2023-06-02
NH-42458 set_transaction_name empty name case

tammy-baylis-swi:
If customers call `set_transaction_name("")` with an empty string, there will now be a warning log and the method will return `False`. I've commented on the [hijacked spec doc](https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3601400601/Swo-SDK+Application+Programming+Interface?focusedCommentId=3620274549) to confirm, and Java library is doing this too: https://github.com/appoptics/solarwinds-apm-java/pull/105/files

Before this change, Python relies on falsiness to not use empty string nor None as transaction name: https://github.com/solarwindscloud/solarwinds-apm-python/blob/main/solarwinds_apm/inbound_metrics_processor.py#L171 So what's exported to SWO doesn't change with this PR.

I've also adjusted log severity of other parts of the method, though I'm not sure what's best with the return of `True` vs `False` and if it's a simple "ignore" or pretty-bad-can't-find-InboundMetricsSpanProcessor.

Any suggestions welcome 😺

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -50,7 +50,7 @@ def set_transaction_name(custom_name: str) -> bool:
         return False
 
     if isinstance(get_tracer_provider(), NoOpTracerProvider):
-        logger.debug(
+        logger.warning(
             "Cannot cache custom transaction name %s because agent not enabled; ignoring",
             custom_name,
         )
@@ -74,7 +74,7 @@ def set_transaction_name(custom_name: str) -> bool:
     entry_trace_id = baggage.get_baggage(INTL_SWO_CURRENT_TRACE_ID)
     entry_span_id = baggage.get_baggage(INTL_SWO_CURRENT_SPAN_ID)
     if not entry_trace_id or not entry_span_id:
-        logger.debug(
+        logger.error(
             "Cannot cache custom transaction name %s because OTel service entry span not started; ignoring",
             custom_name,
         )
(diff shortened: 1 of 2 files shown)


solarwinds_apm/api/__init__.py
```diff
@@ -68,7 +75,7 @@ def set_transaction_name(custom_name: str) -> bool:
     entry_trace_id = baggage.get_baggage(INTL_SWO_CURRENT_TRACE_ID)
     entry_span_id = baggage.get_baggage(INTL_SWO_CURRENT_SPAN_ID)
     if not entry_trace_id or not entry_span_id:
-        logger.debug(
+        logger.error(
```
cheempz, line 78: I think `warning` might be more appropriate, similar to if the customer passed in an invalid transaction name.

solarwinds_apm/api/__init__.py
```diff
@@ -37,14 +37,21 @@ def set_transaction_name(custom_name: str) -> bool:
...
     """
+    if not custom_name:
+        logger.warning(
+            "Cannot set custom transaction name as empty string; ignoring"
+        )
+        return False
+
     if isinstance(get_tracer_provider(), NoOpTracerProvider):
-        logger.debug(
+        logger.warning(
```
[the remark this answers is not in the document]
tammy-baylis-swi: Good call! Don't need too much notification. Addressed in [32bf45e](https://github.com/solarwindscloud/solarwinds-apm-python/pull/162/commits/32bf45effcc669455986fc6cb3a672972d6841d6)

solarwinds_apm/api/__init__.py
tammy-baylis-swi: Thanks, make sense! Addressed in [32bf45e](https://github.com/solarwindscloud/solarwinds-apm-python/pull/162/commits/32bf45effcc669455986fc6cb3a672972d6841d6)

cheempz (approved):

PR #165, opened 2023-06-08
NH-29380 Remove unused clean_hack config

tammy-baylis-swi:
Similar to https://github.com/solarwindscloud/solarwinds-apm-python/pull/163 and https://github.com/solarwindscloud/solarwinds-apm-python/pull/164

This removes `is_grpc_clean_hack_enabled` from APM Python config. It is not currently used and I don't think it's needed since liboboe 12.0.0 (https://swicloud.atlassian.net/browse/NH-28555) but please let me know if I'm wrong! The value is not used to config Reporter nor anything else. This value was copied from AO which I think used it as a hidden option to re-activate a hack (https://github.com/librato/python-appoptics/pull/254).


tammy-baylis-swi:
Closing to include changes in a single PR: https://github.com/solarwindscloud/solarwinds-apm-python/pull/164

PR #179, opened 2023-07-20
NH-46326 Remove unused PackageCloud build workflows

tammy-baylis-swi:
We are not uploading APM Python builds to PackageCloud like we did for AppOptics. So I'm removing its unused workflows (`clean_builds_intermediates` is not used by anyone else) and mentions of it in the smoke test scripts.


cheempz (approved):

PR #182, opened 2023-07-21
NH-46327 APM Python accepts OTEL_SQLCOMMENTER_OPTIONS

tammy-baylis-swi:
Updates APM Python custom distro to read and parse `OTEL_SQLCOMMENTER_OPTIONS` if set by customer and if `OTEL_SQLCOMMENTER_ENABLED=true`. The options are used to init all Otel Python instrumentors that have a sqlcommenting feature. For more details please see ["How the two env vars work" section](https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3704816080/2023-07-24+OTEL+SQLCOMMENTER+OPTIONS+testing#How-the-two-env-vars-work).

This also refactors existing `ApmConfig.convert_to_bool` as a class method so the distro class can also use it. Not ideal so let me know if it should be in its own class or some other way.

Unit tests added and updated.

Tested manually with results starting in ["Results" section](https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3704816080/2023-07-24+OTEL+SQLCOMMENTER+OPTIONS+testing#Results). It seems to work as all only the KVs enabled by options are added to sqlcomments, for example this for Flask with SQLAlchemy and MySQL only has `db_driver` and `traceparent` when all other options are `False`:

`2023-07-24T23:55:39.935228Z	   16 Query	SELECT * FROM city WHERE id='1818' /*db_driver='mysqldb',traceparent='00-6268896697b0a44f98575a72d606120b-263f08ace2b0ddfc-01'*/`

Please let me know if any questions or comments! 😺

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -506,7 +506,7 @@ class SolarWindsApmConfig:
             return
 
         # agent_enabled is special
-        cnf_agent_enabled = self._convert_to_bool(
+        cnf_agent_enabled = self.convert_to_bool(
             cnf_dict.get(_snake_to_camel_case("agent_enabled"))
         )
         if cnf_agent_enabled is not None:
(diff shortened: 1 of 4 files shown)


xuan-cao-swi (approved):
👍

PR #184, opened 2023-07-31
NH-51264 Remove custom naming baggage at propagator.inject

tammy-baylis-swi:


solarwinds_apm/propagator.py
```diff
@@ -108,16 +115,54 @@ def inject(
...
+            )
+            setter.set(carrier, self._BAGGAGE_HEADER_NAME, baggage_header)
+
+    def remove_custom_naming_baggage_header(
+        self,
+        baggage_header: str,
+    ) -> str:
+        """Removes values used for custom naming from baggage header created
+        by upstream W3CBaggagePropagator propagator, if present"""
+        baggage_entries: list[str] = split(_DELIMITER_PATTERN, baggage_header)
```
tammy-baylis-swi, line 135: This split is the same as how W3CBaggagePropgator.extract does it: https://github.com/open-telemetry/opentelemetry-python/blob/b9f31e91a8263bc8effc9abce889d438927d47ec/opentelemetry-api/src/opentelemetry/baggage/propagation/__init__.py#L66

solarwinds_apm/propagator.py
```diff
@@ -108,16 +115,54 @@ def inject(
...
+        baggage_kvs = {}
+        for entry in baggage_entries:
+            try:
+                e_name, e_value = entry.split("=", 1)
```
tammy-baylis-swi, line 139: This is also the same as how W3CBaggagePropagator does it: https://github.com/open-telemetry/opentelemetry-python/blob/b9f31e91a8263bc8effc9abce889d438927d47ec/opentelemetry-api/src/opentelemetry/baggage/propagation/__init__.py#L85

solarwinds_apm/propagator.py
tammy-baylis-swi: This and the other `split` don't have to be this way of course. Open to suggestions if any.

tammy-baylis-swi:
> Though I wonder if it'd be simpler if we used a single baggage key?

Could do! I'll do it in a separate PR.

PR #185, opened 2023-08-01
NH-52515 set transaction name, sampled or not

tammy-baylis-swi:
Custom span processor now always caches (custom) transaction name, even if not sampled. This is so metrics are still reported with those transaction names if sampling decision is record-only.

Tested visually with local test collector, with APM Python temporarily set to always `return Decision.RECORD_ONLY`. Request was made to Django A test app's `home_a/` route including a call to API `set_transaction_name`. We see that before the change it's not using the custom name.

Now:
```
    "histograms": [
        {
            "name": "TransactionResponseTime",
            "value": "<long_unique_value>",
            "tags": {
                "TransactionName": "custom-name-django-b-root"
            }
        }
    ]
```

Before:
```
    "histograms": [
        {
            "name": "TransactionResponseTime",
            "value": "<long_unique_value>",
            "tags": {
                "TransactionName": "home_a/"
            }
        }
    ]
```

### DIFF
--- a/solarwinds_apm/inbound_metrics_processor.py
+++ b/solarwinds_apm/inbound_metrics_processor.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Optional, Tuple
 from opentelemetry import baggage, context
 from opentelemetry.sdk.trace import SpanProcessor
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import SpanKind, StatusCode, TraceFlags
+from opentelemetry.trace import SpanKind, StatusCode
 
 from solarwinds_apm.apm_constants import (
     INTL_SWO_CURRENT_SPAN_ID,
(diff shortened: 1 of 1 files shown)


tammy-baylis-swi:
> The change in this PR I think only adds to the manager after it's already been used in the Processor (i.e. for use with metrics)

I think you are right!!

> In fact, I think it's always set already, but only ever cleared by the exporter, and therefore causing a memory leak?

Set by API method `set_transaction_name` when called (and with this PR always by `on_end` 🙊). Yes, only cleared by the exporter.

> Based on my reading, the apm_txname_manager is queried above this, [in line 90](https://github.com/solarwindscloud/solarwinds-apm-python/pull/185/files#diff-4f89668e1d6f48c51c37cc2e3c7bb6fffaae4347602577eb56721aa0f2188d75R90). In fact, the set_transaction_name also sets this.

`calculate_custom_transaction_name` does a lookup for what might have been set by `set_transaction_name`, but doesn't do an additional assignment. Exporter does the same, then also clears.

I'm going to do a fresh test of this tomorrow because what I pasted into the PR description is not making sense. What I'm thinking is that the current code is already using custom transaction names (via `set_transaction_name`) if set for metrics, whether sampled or not. Memory leak would be introduced if this PR is merged; not sure if there is currently a leak.

tammy-baylis-swi:
> I think there's a leak if a Span is started as record-only, and a user sets the custom transaction name. If custom transaction names are only cleared in the exporter, then this one is never cleared, right?

Yess!! 🔥 

Going to close this and do a different approach.

PR #186, opened 2023-08-01
NH-51264 entry span and trace id as one baggage kv

tammy-baylis-swi:


solarwinds_apm/inbound_metrics_processor.py
```diff
@@ -132,7 +128,7 @@ def on_end(self, span: "ReadableSpan") -> None:
         if span.context.trace_flags == TraceFlags.SAMPLED:
             # Cache txn_name for span export
```
tammy-baylis-swi, line 129: This is going to be addressed in a separate PR for NH-52515.

raphael-theriault-swi (approved):

PR #196, opened 2023-09-08
Update CODEOWNERS

cheempz:


tammy-baylis-swi (approved):
Lgtm, we don't need the other since this is public.

PR #197, opened 2023-09-11
NH-55680 Test agent for lambda layer poc

tammy-baylis-swi:
Always initialize c-lib reporter as `ssl`. This is needed for direct export of OTLP traces, metrics from APM Python running in Lambda environment to SWO/chainsaw.


tammy-baylis-swi:
Closing because we will not need Reporter init in lambda environment if we stick to OTLP export.

PR #200, opened 2023-09-21
NH-60984 Add _configure_metrics_exporter if experimental flag set

tammy-baylis-swi:
This makes the custom distro load the entry points as configured by customers: `OTEL_METRICS_EXPORTER`, `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, `OTEL_EXPORTER_OTLP_METRICS_HEADERS`, etc.

It is similar to the setup that allows customer to configure multiple `OTEL_TRACES_EXPORTER`.

For now, the ability to configure metrics export is only allowed when the experimental flag is also set as True. It can be the environment variable `SW_APM_EXPERIMENTAL_OTEL_COLLECTOR` or in the JSON config file as

```json
  "experimental": {
      "otel_collector": true,
  }
```

What this looks like in action: https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3788865644/OTLP+Metrics+export+via+APM+Python


solarwinds_apm/apm_constants.py
```diff
@@ -17,6 +17,7 @@
 INTL_SWO_X_OPTIONS_KEY = "sw_xtraceoptions"
 INTL_SWO_X_OPTIONS_RESPONSE_KEY = "xtrace_options_response"
 INTL_SWO_DEFAULT_TRACES_EXPORTER = "solarwinds_exporter"
+INTL_SWO_DEFAULT_METRICS_EXPORTER = "otlp_proto_grpc"
```
tammy-baylis-swi, line 20: I'm wondering whose responsibility it should be to install `opentelemetry-exporter-otlp-proto-grpc`: APM Python or the user. If we do make `OTLPMetricExporter` a default that exports to SWO `otel.collector` by default, it should be APM Python. If not, the user perhaps?

cheempz: Oh hmm good question which I completely ignored with @raphael-theriault-swi's https://github.com/solarwindscloud/solarwinds-apm-js/pull/98. For non-lambda use case we're not likely to support exporting via OTLP right away, so in my mind bundling these exporters would be part of the custom lambda layer build and our regular distribution should not install them.

tammy-baylis-swi: Sounds good! Removed this constant and there won't be a default metrics exporter at least for now. `opentelemetry-exporter-otlp-proto-grpc` will be part of the Lambda layer, else users need to install it in their own environment.

cheempz (approved):
LGTM, thanks @tammy-baylis-swi! We had a discussion where @raphael-theriault-swi also thought about calling these `SW_APM_DEV_*` rather than `SW_APM_EXPERIMENTAL_*` to extra reinforce that these features are really under developement, but let's just leave as-is since we got both JS and Python using the experimental wording and IMHO it's understood the same way :)

tammy-baylis-swi:
Thank you @cheempz ! May I please have a fresh approval? I just resolved some conflicts.

cheempz (approved):
👍

PR #204, opened 2023-10-10
NH-63906 Add Debian 11 install tests

tammy-baylis-swi:
Add Debian 11 install tests for on local and in GH workflow.

GH test run here: https://github.com/solarwindscloud/solarwinds-apm-python/actions/runs/6474729597/job/17580341321


cheempz (approved):
LGTM!

PR #209, opened 2023-10-30
NH-65933 Fix workflow yaml syntax

tammy-baylis-swi:
Fix workflow yaml syntax so I can publish, oops.


cheempz (approved):
Yay indentation ;)

PR #216, opened 2023-11-08
NH-65069 Add lambda layer build-publish workflows

tammy-baylis-swi:
First PR to start adding APM Python lambda layer build-publish workflows. It would be great to get this on `main` for the first time to start test runs on GH then fix and enhance.

Adds 3 workflows:
1. `build_publish_lambda_layer_x86_64`
2. `build_publish_lambda_layer_aarch`
3. `publish_lambda_layer` which is called by each of above

Some key features:
* The `build_publish_*` flows are manually triggered for now, assuming called on `main` and not multiple times.
* Zip archive is generated by `make aws-lambda` target, same as local builds.
* AWS CLI is used to publish to our staging account. I did a manual test of the command on my local (with a wrong name) to layer `solarwinds-apm-python_0_18_0_aarch64`
* The published ARNs are supposed to be `arn:aws:lambda:<region>:<account>:layer:solarwinds-apm-python-<arm64|x86_64>-0_18_0:1` at least for now.

I will need some help adding a new secret `STAGING_LAMBDA_ROLE_ARN`. I'm going to refine naming and versioning of APM Python layer in a future PR. I will probably also add a single workflow to trigger both arch builds.

### DIFF
--- a/.github/actions/package_lambda_solarwinds_apm_aarch64/action.yaml
+++ b/.github/actions/package_lambda_solarwinds_apm_aarch64/action.yaml
@@ -1,0 +1,18 @@
+# © 2023 SolarWinds Worldwide, LLC. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at:http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


raphael-theriault-swi:
Could you trigger the workflows one this branch to see an example run ?

tammy-baylis-swi:
> Could you trigger the workflows one this branch to see an example run ?

Right, I could have done this with a temp on-PR trigger!

I also just realized I want to run the aarch builds on an EC2 runner. I'm going to close this one without merging and put in a new PR later.

PR #217, opened 2023-11-10
NH-65713 Adjust exporter defaults and restrictions

tammy-baylis-swi:
### Before

Whether in lambda or outside:
1. No default for `OTEL_METRICS_EXPORTER`
2. `OTEL_TRACES_EXPORTER` defaulted to `solarwinds_exporter`
3. If `OTEL_TRACES_EXPORTER` configured, required to include `solarwinds_exporter` else agent disabled

### After

In lambda:
1. `OTEL_METRICS_EXPORTER` defaults to `otlp_proto_http`
   * (will change in https://swicloud.atlassian.net/browse/NH-66126)
2. `OTEL_TRACES_EXPORTER` defaults to `otlp_proto_http`

Out of lambda:
1. No default for `OTEL_METRICS_EXPORTER`. If `experimental` flag is set, customers must configure metrics exporters else none will initialize.
2. `OTEL_TRACES_EXPORTER` defaults to `solarwinds_exporter`
   * (same as before, to not break)

Note: Metrics exporters will only be initialized if `experimental` flag is set, for both scenarios.

----


cheempz (approved):

PR #215, opened 2023-11-08
NH-65069 AWS lambda layer includes py37-311 extensions

tammy-baylis-swi:
Updates build target so that APM Python AWS lambda layer includes extensions for Python 3.7, 3.8, 3.9, 3.10, 3.11. For example, contents of `extension/` after build on aarch64:

```
drwxr-xr-x  12 tammy.baylis  staff   384B  8 Nov 10:31 .
drwxr-xr-x  26 tammy.baylis  staff   832B  8 Nov 10:31 ..
-rw-r--r--   1 tammy.baylis  staff     7B  8 Nov 10:31 VERSION
-rw-r--r--   1 tammy.baylis  staff   583B  8 Nov 10:31 __init__.py
-rwxr-xr-x   1 tammy.baylis  staff   1.8M  8 Nov 10:31 _oboe.cpython-310-aarch64-linux-gnu.so
-rwxr-xr-x   1 tammy.baylis  staff   1.9M  8 Nov 10:31 _oboe.cpython-311-aarch64-linux-gnu.so
-rwxr-xr-x   1 tammy.baylis  staff   1.8M  8 Nov 10:31 _oboe.cpython-37m-aarch64-linux-gnu.so
-rwxr-xr-x   1 tammy.baylis  staff   1.8M  8 Nov 10:31 _oboe.cpython-38-aarch64-linux-gnu.so
-rwxr-xr-x   1 tammy.baylis  staff   1.8M  8 Nov 10:31 _oboe.cpython-39-aarch64-linux-gnu.so
drwxr-xr-x   4 tammy.baylis  staff   128B  8 Nov 10:31 bson
-rw-r--r--   1 tammy.baylis  staff   2.4M  8 Nov 10:31 liboboe.so
-rw-r--r--   1 tammy.baylis  staff    14K  8 Nov 10:31 oboe.py
```

Resulting zip archive is 11 MB, which was build without this update to shave off 0.8 MB: https://github.com/solarwindscloud/solarwinds-apm-python/pull/214

PR #221, opened 2023-11-11
NH-65069 Add APM Python lambda workflows

tammy-baylis-swi:
Adds new workflow `build_publish_lambda_layer` to publish APM Python on manual dispatch as

`arn:aws:lambda:<region>:<staging_account>:layer:solarwinds-apm-python-<x86_64|arm64>-<apm_python_version>:<layer_version>`

Example: `arn:aws:lambda:us-east-1:<staging_account>:layer:solarwinds-apm-python-arm64-0_18_0:1`

Layer version simply increments with every successful publish for now. Not sure if we want to add "APM Python layer" version in addition to APM Python version and AWS layer version.

Last successful run: https://github.com/solarwinds/apm-python/actions/runs/6831243412

which resulted in publish of these two layers, available in multiple regions:
* [solarwinds-apm-python-x86_64-0_18_0:1](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/layers/solarwinds-apm-python-x86_64-0_18_0/versions/3?tab=versions)
* [solarwinds-apm-python-arm64-0_18_0:1](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/layers/solarwinds-apm-python-arm64-0_18_0/versions/2?tab=versions)

This only works when temporarily using c-lib 14 ahead of its release (commit [96d3223](https://github.com/solarwinds/apm-python/pull/221/commits/96d32233dc7621c3f533748fad193ce24971388a), now reverted), which is not included in this PR.


.github/workflows/build_publish_lambda_layer_aarch64.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+  publish-layer:
+    uses: ./.github/workflows/publish_lambda_layer.yaml
+    needs:
+      - build_layer_aarch64
+    with:
+      artifact-name: solarwinds_apm_lambda.zip
+      component-version: ${{ needs.build_layer_aarch64.outputs.SW_APM_VERSION }}
+      architecture: arm64
+      runtimes: "python3.8 python3.9 python3.10 python3.11"
```
tammy-baylis-swi, line 95: Two notes:

(1)
On this branch, only py3.7 and py3.8 actually have extensions built. Support for 3.9+ will be added in https://github.com/solarwinds/apm-python/pull/215 i.e. I'm prematurely adding them here.

(2)
py3.7 is not compatible with arm64, but it is with x86_64: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

.github/workflows/publish_lambda_layer.yaml
```diff
@@ -0,0 +1,99 @@
...
+        required: true
+        type: string
+      runtimes:
+        description: 'Space-delimited list of compatible runtimes'
+        required: true
+        type: string
+
+permissions:
+  id-token: write
+  contents: read
```
cheempz, line 31: Are these permissions actually needed?

.github/workflows/publish_lambda_layer.yaml
```diff
@@ -0,0 +1,99 @@
...
+          LAYER_NAME=$LAYER_NAME-$COMPONENT_VERSION_UNDERSCORES
+          echo "LAYER_NAME=$LAYER_NAME" >> $GITHUB_ENV
+          
+          echo GITHUB_ENV:
+          cat $GITHUB_ENV
+
+      - name: Download built layer
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact-name }}
```
cheempz, line 80: In each of the build-for-<architecture> workflows, the artifact is uploaded with the same name `solarwinds_apm_lambda.zip`.  I'm wondering if in this publish step, we're guaranteed to be using the correct one, i.e. would the aarch64 artifact have replaced the one built for x86 while we're publishing for x86?

From quick read of about [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) it seems this artifact might be shared since these build-for-<architecture> workflows are part of the main "caller" workflow.  It may be safer to just name the artifacts with architecture appended?

.github/workflows/publish_lambda_layer.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.LAMBDA_PUBLISHER_STAGING_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.LAMBDA_PUBLISHER_STAGING_SECRET_ACCESS_KEY }}
+          aws-region: ${{ matrix.aws_region }}
+
+      - name: Publish Lambda Layer
+        run: |
+          LAYER_ARN=$(
```
cheempz, line 90: We're capturing the resulting LAYER_ARN (nice!) but not sure where it's used--and this means an error exit code from `aws lambda` command will probably not cause the workflow to fail.

cheempz (commented):
Big thank you for putting this together, and the helpful workflow run to look at @tammy-baylis-swi! I left a few comments :)

.github/workflows/publish_lambda_layer.yaml
tammy-baylis-swi: Nope not needed 👍 Removed in [0b10ae7](https://github.com/solarwinds/apm-python/pull/221/commits/0b10ae7dab693b3634e5913c1619dd95146b4e55), tested in https://github.com/solarwinds/apm-python/actions/runs/6879510772/job/18711579223?pr=221

.github/workflows/publish_lambda_layer.yaml
tammy-baylis-swi: Ah yes good point! Removed the capture in [666370a](https://github.com/solarwinds/apm-python/pull/221/commits/666370a5e422bf1402774372ec397f9f5eba8fbc).

We could use it like Otel does to print out the ARN in the workflow https://github.com/open-telemetry/opentelemetry-lambda/blob/cc5a0b6b3c24f3e4dff7a0e960ee87df83ee416e/.github/workflows/layer-publish.yml#L101

`echo "::notice ::$LAYER_ARN component-version=${{ inputs.component-version }}"`

They haven't done a publish of any layer in 4+ months so there's no example logs available.

.github/workflows/publish_lambda_layer.yaml
tammy-baylis-swi: Good call! Changed in workflows and lambda target in https://github.com/solarwinds/apm-python/pull/221/commits/aa42a89813cca8f258b86c613a3b417bed72553b

tammy-baylis-swi:
Thank you @cheempz for those suggestions! Please let me know what you think this time.

cheempz (approved):
LGTM, thanks for the revisit @tammy-baylis-swi!

PR #223, opened 2023-11-15
NH-64716 First steps to integrate oboe settings API

tammy-baylis-swi:
### Summary

1. Adds setting of `ApmConfig.oboe_api` as c-lib OboeAPI if is_lambda and agent_enabled.
2. Else new no-op OboeAPI is used.
3. Sampler uses the OboeAPI it's given for making decision, if is_lambda.
4. Updates MeterManager to init 8 ObservableGauges for sampling metrics, using OboeAPI it's given.

### Caveat

This works with c-lib 14 from staging i.e. [1034e69](https://github.com/solarwinds/apm-python/pull/223/commits/1034e692d5773eb8641b06714ec5132e0dbd2e3d). If c-lib < 14 is used and `is_lambda`, agent will be disabled (see [this message](https://github.com/solarwinds/apm-python/pull/223#discussion_r1394782369)). Since I've reverted this commit, tox tests now fail. They last passed at [5069826](https://github.com/solarwinds/apm-python/pull/223/commits/50698262d8d132719a77470fd1adbf7d1a585b00) before the revert. Not sure if this is the correct way to proceed.

### Manual testing

Contents can be packaged into a usable lambda layer that, with custom otel-collector layer, can instrument a function and export metrics ([example](https://my.na-02.cloud.solarwinds.com/158363224169902080/entities/services/e-1661592723085803520/service-metrics?duration=803&endTime=1700176403)) and traces ([example](https://my.na-02.cloud.solarwinds.com/158363224169902080/traces/7FDAF66681B9C00C1B1A9FB6118CDECB/9929B788835518B3/details/breakdown?perspective=requests&serviceId=e-1661592723085803520&sort_by_breakdown=spanLayer&sort_order_breakdown=ascending)).

Contents work the same as before for locally-instrumented app to export traces and metrics by apm-proto.

Please see doc for more details: https://swicloud.atlassian.net/wiki/spaces/NIT/pages/3881501305/2023-11-16+OboeAPI+integration

### Housekeeping notes

I am adding Configurator tests separately in https://swicloud.atlassian.net/browse/NH-67051

Not sure what else I could do to make the MeterManager tests more meaningful.

Awaiting further input regarding use of `sw.data.module`

Still lots of stuff to do so I've added some Todos with tickies.

----


solarwinds_apm/sampler.py
```diff
@@ -143,43 +140,8 @@ def calculate_liboboe_decision(
         xtraceoptions: Optional[XTraceOptions] = None,
     ) -> dict:
         """Calculates oboe trace decision based on parent span context and APM config."""
-        tracestring = None
```
tammy-baylis-swi, line 146: The main changes of this function are (1) moving the `if` of the `if-else` for lambda to the top and (2) actually using OboeAPI now.

solarwinds_apm/sampler.py
```diff
@@ -192,21 +154,40 @@ def calculate_liboboe_decision(
...
-                0,
-                0.0,
-                0.0,
-                0,
-                0,
-                "",
-                "",
-                0,
-            )
+            ) = self.oboe_settings_api.getTracingDecision()
```
raphael-theriault-swi, line 157: Do the arguments to this call not matter ? Actually curious if passing it the same thing as the non-lambda version vs passing it nothing makes any difference or if I should be doing it differently on my side of things

raphael-theriault-swi (approved):
Looks good but I had a question

solarwinds_apm/sampler.py
tammy-baylis-swi: Ah good question. Tried it now and yes, this method does not explode if passed all the same non-lambda args (though it's a drop decision). Works if no args at all of course.

It does error out if I try other combinations of args:`Wrong number or type of arguments for overloaded function`

PR #225, opened 2023-11-20
NH-65713 Add configurator trace exporter tests

tammy-baylis-swi:
Adds unit tests for configurator's `_configure_traces_exporter` function. Includes `fixtures/` that I'll re-use for other configurator tests in next PRs.

Does a teeny src code change so that the `Failed to load` message doesn't print out info originating from env var anymore (the ApmConfig agent_enabled method does it this way).

Last test run at [4b35b69](https://github.com/solarwinds/apm-python/pull/225/commits/4b35b69f049b6773b7d553e0faf558d042dffd96) passed with c-lib 14. Then fails again with revert, like https://github.com/solarwinds/apm-python/pull/224


tests/unit/test_configurator/test_configurator_traces_exporter.py
```diff
@@ -0,0 +1,188 @@
...
+        test_configurator = configurator.SolarWindsConfigurator()
+        test_configurator._configure_traces_exporter(
+            mock_reporter,
+            mock_txn_name_manager,
+            mock_fwkv_manager,
+            mock_apmconfig_enabled,
+        )
+        mock_bsprocessor.assert_not_called()
+        mock_get_tracer_provider.assert_not_called()
+        mock_add_span_processor.assert_not_called()
```
cheempz, line 79: For this `traces_exporter_none` test, does it mean if `OTEL_TRACES_EXPORTER` env var is not set, our custom distro does not configure the solarwinds_exporter either?

cheempz (approved):
LGTM, great to see the foundations you've added for testing @tammy-baylis-swi! Left a question that may be my misreading, but it seems if exporter isn't explicitly set we should default to the solarwinds one?

tests/unit/test_configurator/test_configurator_traces_exporter.py
tammy-baylis-swi: Ooh yes good question.

For `_configure_traces_exporter` in complete isolation, yes it could be that `OTEL_TRACES_EXPORTER` is None so this test is for that scenario.

For `_configure_traces_exporter` running at distro startup, that scenario shouldn't happen. [SolarWindsDistro._configure](https://github.com/solarwinds/apm-python/blob/b97fff717a84e9bd9be5efd22f882f35f5ccac05/solarwinds_apm/distro.py#L45-L53) is called first to set env var [defaults](https://github.com/solarwinds/apm-python/blob/b97fff717a84e9bd9be5efd22f882f35f5ccac05/solarwinds_apm/apm_constants.py#L19-L23). Outside lambda the default is the SW exporter. Then [SolarWindsConfigurator._configure](https://github.com/solarwinds/apm-python/blob/b97fff717a84e9bd9be5efd22f882f35f5ccac05/solarwinds_apm/configurator.py#L80) is called to init the exporters etc. If `OTEL_TRACES_EXPORTER` wasn't set by user, then default from distro is used instead.

cheempz: Ah gotcha, thanks for the pointers :)

PR #231, opened 2023-11-22
NH-65713 Add configurator metrics exporter tests

tammy-baylis-swi:
Add Configurator's metrics exporter setup method tests (similar to trace exporter setup testing in https://github.com/solarwinds/apm-python/pull/225).

I've renamed some existing fixtures to try to make CodeQL happier with their imports.

Also does a teeny src code change so that the Failed to load message doesn't print out info originating from env var anymore (also like other PR).

Last test run at [2670491](https://github.com/solarwinds/apm-python/pull/231/commits/26704913fdab37194b98177ac6b5d6b023ba52d4) passed with c-lib 14. Then fails again only after I revert.


cheempz (approved):

PR #232, opened 2023-11-22
NH-67783 Update APM Python package classifiers and metadata

tammy-baylis-swi:
Update APM Python package classifiers and metadata because GA, py3.11 has been supported a while, and updated url.


cheempz (approved):
Nice!

PR #233, opened 2023-11-22
NH-67051 Add span processor config tests

tammy-baylis-swi:
Add span processor configurator unit tests.


cheempz (approved):
LGTM, thanks @tammy-baylis-swi :)

PR #235, opened 2023-11-22
NH-67051 Add mixed exporter config tests

tammy-baylis-swi:
Adds more configurator tests for exporter config, where `OTEL_TRACES_EXPORTER` or `OTEL_METRICS_EXPORTER` has more than one name.

Passes when c-lib 14 ([88128e5](https://github.com/solarwinds/apm-python/pull/235/commits/88128e5319680951a9096462990dcfedadc0e776))


cheempz (approved):
👍

PR #236, opened 2023-11-23
NH-67051 Add sampler configurator tests

tammy-baylis-swi:
Adds Sampler configurator methods tests in `test_configurator_sampler.py`.

Updates the `Resource` and `Trace` fixtures, and the other tests that use them.

Tests pass with c-lib 14 before my last revert.

I will set up `conftest` in a next PR!


cheempz (approved):
LGTM, there are a few CodeQL `Unused local variable` findings which if have merit feel free to address in follow-on PR.

PR #237, opened 2023-11-23
NH67051 Rearrange Configurator test fixtures

tammy-baylis-swi:
Rearrange several Configurator test fixtures because they should be in pytest `conftest.py` instead of imported for use as side-effects. This should get rid of several CodeQL unused import detections.

Also updates the trace, metrics, and Resource fixtures to each return a single obj so it's easier to make changes later.

Tests pass with c-lib 14 ([eb1c0a4](https://github.com/solarwinds/apm-python/pull/237/commits/eb1c0a4f04e17fe7870b1f1b8775a06c12aafa8c))


cheempz (approved):
👍

PR #238, opened 2023-11-23
NH-67051 Add Reporter config tests

tammy-baylis-swi:
Adds Reporter configurator function unit tests, with updated fixtures.

At distro startup, it's the upstream `_configure` method that decides if Reporter should be init'd.


cheempz (approved):

PR #239, opened 2023-11-23
NH-67051 Refactor configurator test fixtures more

tammy-baylis-swi:
Refactors the configurator unit test fixtures some more, before I add even more for next tests.

Switches returns and usage of trace, metrics, resource test fixtures so tests use dot notation instead.

Small change: moves the `MeterProvider` patch out for easier use.


cheempz (approved):
`.` is nicer to read :)

PR #240, opened 2023-11-24
NH-67051 Add report_init_event tests

tammy-baylis-swi:
Adds report_init_event tests!! Tricky mocks for this one.

Tests pass with c-lib 14 ([5d7dcee](https://github.com/solarwinds/apm-python/pull/240/commits/5d7dcee8a41f47459cb66de62ce5727899d29b47))  before revert


tests/unit/test_configurator/conftest.py
```diff
@@ -137,6 +275,40 @@ def mock_extension(mocker):
...
+def mock_extension(mocker):
+    return get_extension_mocks(mocker)
+
+@pytest.fixture(name="mock_extension_status_code_already_init")
+def mock_extension_status_code_already_init(mocker):
+    return get_extension_mocks(mocker, -1)
+
+@pytest.fixture(name="mock_extension_status_code_invalid_protocol")
+def mock_extension_status_code_invalid_protocol(mocker):
+    return get_extension_mocks(mocker, 2)
```
cheempz, line 288: Great addition of these different scenarios "ok", "already_init" and "invalid_protocol"!  It's pointing out something a little subtle--we're checking the reporter initialization status as part of [reporting the Init event](https://github.com/solarwinds/apm-python/blob/51374972115712b616c8e19302e7b2f679584ed4/solarwinds_apm/configurator.py#L537), and actually ignoring the return code from the report Init event `sendStatus` call.

tammy-baylis-swi:
> One future possibility is to have an explicit check of the "extension" (whether full oboe with reporter, or minimal Lambda api) init/loading status rather than as part of sending Init event.

Thank you @cheempz ! New ticket here: https://swicloud.atlassian.net/browse/NH-68251

PR #252, opened 2023-12-13
NH-65061 Add separate ServiceEntryId-storing span processor

tammy-baylis-swi:
Factors out a new Service Entry trace-span ID-storing span processor, separate from InboundMetricsSpanProcessor.

Also renames an existing test mock after method rename in #251.


raphael-theriault-swi (approved):
Looks good !

PR #258, opened 2023-12-19
NH-68264 Configurator does the only OboeAPI init

tammy-baylis-swi:
Before this change: custom Sampler and MeterManager each init'd their own instances of OboeAPI.

After this change: the Configurator init's one OboeAPI that gets passed to Sampler and MeterManager. I wanted to do this before updating the init call args.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -97,6 +97,7 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
         apm_config = SolarWindsApmConfig()
         # TODO add args
         # https://swicloud.atlassian.net/browse/NH-68264
+        # See also SolarWindsApmConfig._get_extension_components
         oboe_api = apm_config.oboe_api()
 
         # TODO Add experimental trace flag, clean up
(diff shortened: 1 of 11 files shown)


solarwinds_apm/configurator.py
```diff
@@ -87,6 +95,10 @@ def _configure(self, **kwargs: int) -> None:
         apm_txname_manager = SolarWindsTxnNameManager()
         apm_fwkv_manager = SolarWindsFrameworkKvManager()
         apm_config = SolarWindsApmConfig()
+        # TODO add args
+        # https://swicloud.atlassian.net/browse/NH-68264
+        # See also SolarWindsApmConfig._get_extension_components
```
tammy-baylis-swi, line 100: If not agent_enabled or not is_lambda, then this is no-op. Defined [here](https://github.com/solarwinds/apm-python/blob/66a6e60d9d6911a05921b22eb00bfdbfe8173ed3/solarwinds_apm/apm_config.py#L156-L166):

```
agent_enabled T, is_lambda F -> c-lib extension, c-lib Context, no-op settings API
agent_enabled T, is_lambda T -> no-op extension, no-op Context, c-lib settings API
agent_enabled F              -> all no-op
```

raphael-theriault-swi (approved):

PR #259, opened 2023-12-19
NH-68264 Upgrade c-lib 14

tammy-baylis-swi:
Depends on #258 

Upgrades APM Python to c-lib 14 which includes updated logging. APM Python therefore includes these updates:

1. ~~Adds support for SW_APM_LOG_TYPE env var / logType json file configs~~ EDIT This will not be a supported config.
2. Init Reporter with `log_type` (with existing `debug_level`)
3. Init OboeAPI with `log_type` and `debug_level`
4. ~~Deprecate SW_APM_DEBUG_LEVEL: -1. Remove later in https://swicloud.atlassian.net/browse/NH-69517. If customer sets -1 then _update_log_settings logs warning and automatically sets log_type to DISABLED and debug_level to 0 instead.~~ EDIT We will be keeping the ability to set `SW_APM_DEBUG_LEVEL: -1`.
5. `update_log_settings` helper sets internal `log_type` to disabled if debug_level -1, or file if logname (filepath) is not none.
6. Removed those icky conditional imports of OboeAPI
7. Slight change from #258 to init OboeAPI in ApmConfig instead of in Configurator

~~Will be looking closer at log_type: <file> in https://swicloud.atlassian.net/browse/NH-69396~~

I might change the Python logging handler logic in a different PR.


solarwinds_apm/apm_logging.py
```diff
@@ -38,13 +38,46 @@
...
+        return cls.log_types["DISABLED"]
+
+    @classmethod
+    def is_valid_level(cls, level):
+        """Returns True if the provided level is a valid interger representation of log type, False otherwise."""
+        try:
+            level = int(level)
+            return bool(level in list(cls.log_types.values()))
+        except (ValueError, TypeError):
+            return False
```
raphael-theriault-swi, line 69: Wouldn't these make more sense as `default_type`/`disabled_type`/`is_valid_type` ?

solarwinds_apm/apm_logging.py
```diff
@@ -38,13 +38,46 @@
...
-    """Abstract mapping class providing a conversion between solarwinds_apm agent logging level and Python logging module
+    """Abstract mapping class providing a conversion between solarwinds_apm library logging level and Python logging module
     logging levels.
     The solarwinds_apm package has seven different log levels, which are defined in the debug_levels dictionary.
     """
 
     # maps string representation of solarwinds_apm.sw_logging levels to their integer counterpart
+    # TODO no longer allow -1
+    # https://swicloud.atlassian.net/browse/NH-69517
     debug_levels = {
```
raphael-theriault-swi, line 81: I'm pretty sure the available levels and their meanings changed with oboe 14, not sure if you already updated these though

solarwinds_apm/apm_logging.py
tammy-baylis-swi: Yes! Changed in [62624db](https://github.com/solarwinds/apm-python/pull/259/commits/62624db411743d7443e6f07d6165e1e9d49a004d)

solarwinds_apm/apm_logging.py
tammy-baylis-swi: We're going to keep level `-1` as a valid level for customers to config log disabling, though it shouldn't be used internally. If level is `-1` then log type will be disabled by `ApmConfig.update_log_type` and log level should be ignored by oboe.

solarwinds_apm/apm_config.py
```diff
@@ -134,17 +135,29 @@ def __init__(
             self.__config["service_key"],
             self.service_name,
         )
+        self.update_log_type()
+        # update logging level of Python logging logger
+        apm_logging.set_sw_log_level(self.__config["debug_level"])
```
tammy-baylis-swi, line 140: Moved this up here instead of in `_set_config_value` for clarity and so only called once.

solarwinds_apm/apm_logging.py
```diff
@@ -38,10 +38,45 @@
...
 
 
+class ApmLoggingType:
+    """Mapping of supported solarwinds_apm library log types"""
+
+    log_types = {
+        "STDERR": 0,
+        "FILE": 2,
+        "DISABLED": 4,
+    }
```
tammy-baylis-swi, line 48: I've not included stdout nor null because there currently are no cases where APM Python would need to set them. But if too strict I can add them.

raphael-theriault-swi (approved):
Thanks for the changes !

PR #260, opened 2024-01-04
NH-66478 Remove unused Mend/whitesource repo config

tammy-baylis-swi:
Remove Mend/whitesource repo-level config because no longer being used.


cheempz (approved):
👍

PR #261, opened 2024-01-05
NH-66478 Add dependabot config

tammy-baylis-swi:
Adds dependabot config for APM Python to do dependency updates. See also comment: https://swicloud.atlassian.net/browse/NH-66478?focusedCommentId=2584965


cheempz (approved):
👍

PR #270, opened 2024-01-05
NH-66478 Update dependabot config

tammy-baylis-swi:
Updates pip ecosystem groups for Dependabot dependency checks to `otel-core`, `otel-instrumentors`, `otel-utils`, `misc` for PR clarity. Also reverts pip ecosystem checks back to `weekly` so PRs don't happen on weekend.


cheempz (approved):
LGTM :)

PR #278, opened 2024-01-17
NH-70103 Update lambda publishing

tammy-baylis-swi:
Adds `Make Lambda layer public` step and switches auth to use arn.


cheempz (approved):

PR #282, opened 2024-01-19
NH-70984 Fix release PR action

tammy-baylis-swi:
This changes the create_release_pr workflow so that a bash script is used to call `gh api` for the commits to be trusted. I've trimmed down `solarwinds_apm/version.py` to be a single line for easier file changes; I hope not having the header is ok.

Outcome is same as before: the workflow is manually triggered to create a draft PR that needs manual approval. Example PR generated from this action: https://github.com/solarwinds/apm-python/pull/283


cheempz (approved):
Neat!

PR #284, opened 2024-01-20
NH-70994 Remove sw.nonce from metrics attributes

tammy-baylis-swi:
Remove `sw.nonce` metrics attribute setting by APM Python, as ingestion does that now.


cheempz (approved):
👍

PR #286, opened 2024-01-22
PoC Fix metrics

tammy-baylis-swi:
WIP / PoC only

[d105303](https://github.com/solarwinds/apm-python/pull/286/commits/d105303896a104ea126557e227b5aa23bb4eacf4) moves the order of metrics exporter init earlier so that preferred temporality/aggregation is set before meters are created (so it's more like [this Otel example](https://github.com/open-telemetry/opentelemetry-python/blob/main/docs/examples/metrics/reader/preferred_aggregation.py)). It also moves MeterManager init into the MetricsSpanProcessor because no other configured Otel components need the meters.


tammy-baylis-swi:
Closing this because implemented with tests in several other PRs (#289 onwards)

PR #289, opened 2024-01-25
NH-70998 Init metrics exporters before meters

tammy-baylis-swi:
This moves the order of metrics exporter init earlier so that preferred temporality/aggregation can later be set before meters are created (so it's more like [this Otel example](https://github.com/open-telemetry/opentelemetry-python/blob/main/docs/examples/metrics/reader/preferred_aggregation.py) and our metrics config behaves as expected).

It also moves MeterManager init into the MetricsSpanProcessor because no other configured Otel components need the meters. Therefore adds `oboe_api` arg to `OTLPMetricsSpanProcessor` init, and removes `apm_meters` arg.


cheempz (approved):

PR #291, opened 2024-01-29
NH-71430 Add log of runtime, notify contributors of unsupported env

tammy-baylis-swi:
Adds more logging to the custom Distro class, which is the first entry point loaded by `opentelemetry-instrument` if APM Python is installed. These log first and will always be in console because of how we configure Boost.log, even if `SW_APM_LOG_FILEPATH` is set. Example:

```
[ solarwinds_apm.distro DEBUG    p#1.281472858361888] SolarWinds APM Python 1.2.0.0
[ solarwinds_apm.distro DEBUG    p#1.281472858361888] Python 3.7.17
[ solarwinds_apm.distro WARNING  p#1.281472858361888] Deprecation: Python 3.7.17 is at end-of-life and support by APM Python will be dropped in a future release. Please upgrade.
[ solarwinds_apm.distro DEBUG    p#1.281472858361888] OpenTelemetry 1.20.0/0.41b0
```

I realize I'm using both `platform` and `sys` stdlib modules; just seemed easier to get the above string formatting and also get `major` and `minor` versions.

A different update included in this PR is duplication of some `project` metadata in the `setup.py` setup function. This is so that contributors installing locally, who might be using e.g. Python 3.6 with an old setuptools version, will get a more informative error e.g. `ERROR: Package 'solarwinds-apm' requires a different Python: 3.6.15 not in '>=3.7'`

Let me know if the logs could use any changes!


cleverchuk (dismissed):
LGTM

cheempz:
One comment :) -- I think it would be helpful if these messages are logged at a level that they would by default show up (is it INFO?). The customer can set a higher level if they really didn't want to see them.

tammy-baylis-swi:
> One comment :) -- I think it would be helpful if these messages are logged at a level that they would by default show up (is it INFO?). The customer can set a higher level if they really didn't want to see them.

@cheempz I've upgraded them to level `INFO` in [6f2c9be](https://github.com/solarwinds/apm-python/pull/291/commits/6f2c9be10f3ca625c767781b9f77d73fe30699ff). As we chatted, they won't show up by default because `SW_APM_DEBUG_LEVEL` default is `WARNING` (all levels [here](https://docs.python.org/3/library/logging.html#logging-levels)). I think it's best to have these as `INFO` for the English meaning, and I don't want to change default level at this time.

Let me know what you think!

PR #295, opened 2024-02-01
NH-70998 Fix no-op function signature

tammy-baylis-swi:
Fix clib API getTracingDecision no-op function signature, to stop TypeError if lambda and no-op.


cheempz (approved):
👍

PR #296, opened 2024-02-01
NH-70998 Set histogram preferred temporality delta

tammy-baylis-swi:
Depends on https://github.com/solarwinds/apm-python/pull/289. EDIT: that's done thanks!

Sets preference for all histograms to report using delta aggregation, including `response_time`.

The entry point `load` is called for any metrics exporters specified by `OTEL_METRICS_EXPORTER`. Otel Python's otlp metrics exporters for [http](https://github.com/open-telemetry/opentelemetry-python/blob/1061d96562990b1c2b596ace87239e7f9616589a/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py#L103) and [grpc](https://github.com/open-telemetry/opentelemetry-python/blob/1061d96562990b1c2b596ace87239e7f9616589a/exporter/opentelemetry-exporter-otlp-proto-grpc/src/opentelemetry/exporter/otlp/proto/grpc/metric_exporter/__init__.py#L101) both accept optional `preferred_temporality`. I don't think there are others in Otel Python. If someone configures a custom metrics exporter, it needs to also accept temporality or general **kwargs, else this might error.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -32,8 +32,6 @@ from opentelemetry.metrics import set_meter_provider
 from opentelemetry.propagate import set_global_textmap
 from opentelemetry.propagators.composite import CompositePropagator
 from opentelemetry.sdk._configuration import _OTelSDKConfigurator
-from opentelemetry.sdk.metrics import MeterProvider
-from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
 from opentelemetry.sdk.metrics import Histogram, MeterProvider
 from opentelemetry.sdk.metrics.export import (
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM, thanks @tammy-baylis-swi for highlighting the impact of this change. 

From quick read of info online, the temporality to choose comes down to what the backend supports--since we're building that out in SWO based on delta, and metrics exporting is still behind the experimental setting, it seems fine to go with setting it on all exporters for now.

In the future we may want more flexibility which might only be possible by creating a SWO-specific exporter.

PR #297, opened 2024-02-01
NH-70998 PeriodicExportingMetricReader with export_interval infinity

tammy-baylis-swi:
We init [PeriodicExportingMetricReader](https://github.com/open-telemetry/opentelemetry-python/blob/1061d96562990b1c2b596ace87239e7f9616589a/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/export/__init__.py#L420-L428) with export interval infinity so reading is no longer periodic.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -8,6 +8,7 @@
 
 import importlib
 import logging
+import math
 import os
 import sys
 import time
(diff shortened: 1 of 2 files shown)


cheempz (approved):
👍 same thought as for the histogram temporality delta, that in the future we may want to limit "explicit reading" to only certain scenarios (like lambda). Just a future mental note :)

PR #298, opened 2024-02-01
NH-70998 OTLP response_time metrics in milliseconds

tammy-baylis-swi:
Fixes OTLP response_time metrics so they're reported in milliseconds (not nanoseconds, as before).

Here is a test otelcol metrics message following a request to a route with a `time.sleep(3)` included:

```
Metric #0
Descriptor:
     -> Name: trace.service.response_time
     -> Description: measures the duration of an inbound HTTP request
     -> Unit: ms
     -> DataType: Histogram
     -> AggregationTemporality: Cumulative
HistogramDataPoints #0
Data point attributes:
     -> sw.is_error: Str(false)
     -> http.status_code: Int(200)
     -> http.method: Str(GET)
     -> sw.transaction: Str(unused)
StartTimestamp: 2024-02-01 18:28:18.418123297 +0000 UTC
Timestamp: 2024-02-01 18:28:18.420461838 +0000 UTC
Count: 1
Sum: 3017.000000
Min: 3017.000000
Max: 3017.000000
```

### DIFF
--- a/solarwinds_apm/trace/base_metrics_processor.py
+++ b/solarwinds_apm/trace/base_metrics_processor.py
@@ -90,9 +90,13 @@ class _SwBaseMetricsProcessor(SpanProcessor):
         self,
         start_time: int,
         end_time: int,
+        time_conversion: int = 1e3,
     ) -> int:
-        """Calculate span time in microseconds (us) using start and end time
-        in nanoseconds (ns). OTel span start/end_time are optional."""
+        """Calculate span time (via time_conversion e.g. 1e3, 1e6)
+        using start and end time in nanoseconds (ns). OTel span
+        start/end_time are optional."""
         if not start_time or not end_time:
             return 0
-        return int((end_time - start_time) // 1e3)
+        ms_start_time = int(start_time // time_conversion)
+        ms_end_time = int(end_time // time_conversion)
+        return ms_end_time - ms_start_time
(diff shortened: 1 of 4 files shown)


solarwinds_apm/trace/base_metrics_processor.py
```diff
@@ -90,9 +90,13 @@ def calculate_span_time(
         self,
         start_time: int,
         end_time: int,
+        time_conversion: int = 1e3,
     ) -> int:
-        """Calculate span time in microseconds (us) using start and end time
-        in nanoseconds (ns). OTel span start/end_time are optional."""
+        """Calculate span time (via time_conversion e.g. 1e3, 1e6)
+        using start and end time in nanoseconds (ns). OTel span
+        start/end_time are optional."""
```
cheempz, line 97: Curious why otel span start/end times are optional?

cheempz (approved):
LGTM @tammy-baylis-swi pending one question.

solarwinds_apm/trace/base_metrics_processor.py
tammy-baylis-swi: They're technically optional for [ReadableSpan](https://github.com/open-telemetry/opentelemetry-python/blob/1061d96562990b1c2b596ace87239e7f9616589a/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L368-L369) in terms of init args. I think auto calculation by the SDK (maybe [this?](https://github.com/open-telemetry/opentelemetry-python/blob/1061d96562990b1c2b596ace87239e7f9616589a/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L880)) is the main use case for either instrumentors or start_as_current_span so I imagine almost always there are start/end times. I don't know when there wouldn't be but might as well catch for that case.

cheempz: Gotcha!

PR #299, opened 2024-02-01
NH-72055 Remove meters for sample_rate, sample_source

tammy-baylis-swi:
As per spec update, removes the OTLP meters (regular and no-op) that report `trace.service.sample_rate` and `trace.service.sample_source`.

Keeps the API no-op methods since extension still has them.


cheempz (approved):
👍

PR #300, opened 2024-02-01
NH-70998 Rename Gauge callback local variables

tammy-baylis-swi:
Cosmetic change to rename ObservableGauge callback functions' local variables. Just for clarity, and noticed these after the c-lib 14.0.2 update.

Does not change for the sample_rate/source callbacks because they'll be removed in https://github.com/solarwinds/apm-python/pull/299


cheempz (approved):
Hehe yes that's clearer ;)

PR #302, opened 2024-02-03
NH-71963 Fix release_pr script

tammy-baylis-swi:
Fix release_pr script to update version.py with better style, for example:

```
__version__ = "1.2.3"

```

instead of

```
__version__ ='1.2.3'
```


cheempz (approved):
LGTM

PR #310, opened 2024-02-09
NH-66126 Exporter defaults based on configured protocol

tammy-baylis-swi:
Before this PR, metrics and traces exporter defaults depended on is_lambda (see https://github.com/solarwinds/apm-python/pull/217).

After this PR, the defaults don't depend on lambda. They instead depend on `OTEL_EXPORTER_OTLP_PROTOCOL` which we require for otlp exporting in any environment.

If Protocol is `http/protobuf`:

1. `OTEL_METRICS_EXPORTER` defaults to otlp_proto_http
2. `OTEL_TRACES_EXPORTER` defaults to otlp_proto_http

If Protocol is `grpc`:

1. `OTEL_METRICS_EXPORTER` defaults to otlp_proto_grpc
2. `OTEL_TRACES_EXPORTER` defaults to otlp_proto_grpc

If Protocol is anything else or unset:

1. No default for `OTEL_METRICS_EXPORTER`. User must configure metrics exporters else none will initialize.
2. `OTEL_TRACES_EXPORTER` defaults to solarwinds_exporter

Note that the above are for default values. Any combination of env vars to specify protocol, exporter, endpoint will still be respected.

This does deviate from the Otel default distro, which uses otlp_proto_grpc by default.

We now start packaging otlp_proto_grpc with our layer to better match Otel's layer and because export by grpc is available and working and the default for Otel Python.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -336,14 +336,18 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                     INTL_SWO_SUPPORT_EMAIL,
                 )
                 raise
-            logger.debug(
-                "Setting trace with BatchSpanProcessor using %s",
-                exporter_name,
-            )
 
             if apm_config.is_lambda:
+                logger.debug(
+                    "Setting trace with SimpleSpanProcessor using %s",
+                    exporter_name,
+                )
                 span_processor = SimpleSpanProcessor(exporter)
             else:
+                logger.debug(
+                    "Setting trace with BatchSpanProcessor using %s",
+                    exporter_name,
+                )
                 span_processor = BatchSpanProcessor(exporter)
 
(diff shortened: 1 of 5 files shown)


solarwinds_apm/configurator.py
```diff
@@ -336,14 +336,18 @@ def _configure_traces_exporter(
...
+                logger.debug(
+                    "Setting trace with SimpleSpanProcessor using %s",
+                    exporter_name,
+                )
                 span_processor = SimpleSpanProcessor(exporter)
             else:
+                logger.debug(
+                    "Setting trace with BatchSpanProcessor using %s",
+                    exporter_name,
+                )
```
tammy-baylis-swi, line 350: Unrelated: updates some misleading debug lines

cheempz (approved):
LGTM, thanks for the nice PR description @tammy-baylis-swi!

PR #312, opened 2024-02-17
NH-71974 Remove ForceFlushSpanProcessor

tammy-baylis-swi:
Removes APM Python `ForceFlushSpanProcessor` because it causes empty otlp metrics messages to export when used with upstream `AwsLambdaInstrumentor`. With that upstream, trace and metrics already get flushed.


cheempz (approved):
👍

PR #320, opened 2024-02-21
NH-70772 Add formatter to rotatingfilehandler

tammy-baylis-swi:
Adds formatter to RotatingFileHandler so that logs in contents of `SW_APM_LOG_FILEPATH` match what the StreamHandler already does. Example logs

```
2024-02-21 17:04:46,652 [ solarwinds_apm.configurator WARNING  p#1.281472885620768] Agent disabled. Not sending init message.
2024-02-21 17:04:49,552 [ solarwinds_apm.api DEBUG    p#1.281472856486304] Cannot cache custom transaction name abc because agent not enabled; ignoring
```


cheempz (approved):
👍

PR #254, opened 2023-12-13
NH-65061 Refactor metrics processors, updated txn name caching, new config restriction

tammy-baylis-swi:
### Before changes

`SolarWindsInboundMetricsSpanProcessor` and `SolarWindsOTLPMetricsSpanProcessor` each did too many things and duplicated some of what the other was doing. The OTLP processor depended on / assumed that the Inbound processor always ran first. Inbound would calculate (custom) transaction name and cache it in `SolarWindsTxnNameManager`, always assuming sw exporter would run after to remove from cache. Big mess.

### After changes

1. Up to 5 span processors are init'd at startup (3 before).
2. New `TxnNameCalculatorProcessor` runs _before_ the metrics processors. It calculates `(trans_name, url_trans)` and caches in `SolarWindsTxnNameManager` by `<trace_id>-<span_id>`.
3. New `TxnNameCleanupProcessor` runs _after_ the metrics processors. It removes `(trans_name, url_trans)` from cache.
4. New `_SwBaseMetricsProcessor` is extended by Inbound and OTLP processors. Contains shared utils like getting `(trans_name, url_trans)` from cache.
5. Inbound metrics processor calculates and caches final oboe-calculated name in `SolarWindsTxnNameManager` by `oboe-<trace_id>-<span_id>`, a different key with a prefix.
6. Solarwinds traces exporter removes oboe names by prefixed key from `SolarWindsTxnNameManager` after addInfo and span export.
7. New restriction: inbound processor only inits if sw traces exporter will be init'd.

### Testing

Added and updated lots of unit tests. Also adds OTLP processor and forceflush processor tests that were missing.

[APM-proto test trace](https://my.na-01.cloud.solarwinds.com/140638900734749696/traces/2C10984BEAAA4CEC88FBFE9D200A9CFC/AA4A98A474332999/details/breakdown/AA4A98A474332999/raw-data?perspective=requests&serviceId=e-1655756171056209920&sort_by_breakdown=spanLayer&sort_order_breakdown=ascending) with custom transaction naming still being applied. [Exported inbound metrics](https://my.na-01.cloud.solarwinds.com/140638900734749696/entities/services/e-1655756171056209920/service-metrics?duration=43859&endTime=1702665659) from same time period.

[...]

PR #330, opened 2024-03-04
NH-69129 Remove todos

tammy-baylis-swi:
Remove some todos because these already no longer log after a [previous PR](https://github.com/solarwinds/apm-python/pull/254).


raphael-theriault-swi (approved):
:tada:

PR #341, opened 2024-03-21
NH-75964 Add lambda production publish workflow

tammy-baylis-swi:
Adds lambda production publish workflow by adding a new `publish-dest` input param to the existing workflow for staging. The input determines whether we use the existing staging publish role or the new role for prod. I've also renamed the reused workflows to make it a bit more clear which one is for dispatch: `Publish APM Python lambda layer`.

Tested [production publish](https://github.com/solarwinds/apm-python/actions/runs/8381326109) resulting in layers ending in `1_6_0_2:1`. Tested [staging publish](https://github.com/solarwinds/apm-python/actions/runs/8381120357) resulting in layers ending in `1_6_0_2:2`. I did these on [another PR](https://github.com/solarwinds/apm-python/pull/340) for the custom build branch and this PR ports over those changes for `main`.

### DIFF
--- a/.github/workflows/build_publish_lambda_layer.yaml
+++ b/.github/workflows/build_publish_lambda_layer.yaml
@@ -8,12 +8,25 @@ name: Publish APM Python lambda layer
 
 on:
   workflow_dispatch:
-
+    inputs:
+      publish-dest:
+        required: true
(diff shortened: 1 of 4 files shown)


cheempz (approved):
LGTM, thanks for the example workflows @tammy-baylis-swi!

PR #344, opened 2024-03-25
NH-76081 Drop Python 3.7 support (lambda only)

tammy-baylis-swi:
Drops Python 3.7 support because [Otel Python 1.23.0/0.44b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.23.0) drops Python 3.7 support. Updates some comments around pylint `deprecated-typing-alias` warnings because Python 3.8 still needs to use the deprecated module for type checks.


tammy-baylis-swi:
Tests will not pass because of some custom build settings. See also ticket comment.

PR #357, opened 2024-04-11
NH-75342 Loggers stop logging handler propagation

tammy-baylis-swi:
Fixes APM log duplication that was happening in these different scenarios:
1. instrumented app code (user side) calls [logging.basicConfig](https://docs.python.org/3/library/logging.html#logging.basicConfig)
2. `OTEL_PYTHON_LOG_CORRELATION` enabled (in and out of lambda), because LoggingInstrumentor [also calls logging.basicConfig](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fdcbbddb6c753e5e9d494ba399a5b4bcab4afc3f/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L147)
3. `OTEL_PYTHON_LOG_CORRELATION` enabled and `SW_APM_LOG_FILEPATH` set

### DIFF
--- a/solarwinds_apm/apm_logging.py
+++ b/solarwinds_apm/apm_logging.py
@@ -163,10 +163,10 @@ def _get_logger():
 
     _logger.setLevel(ApmLoggingLevel.logging_map[log_level])
 
+    _logger.propagate = False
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM!

PR #375, opened 2024-06-03
NH-79326 Clean up layer builds after Otel upgrade

tammy-baylis-swi:
Cleans up lambda layer builds following Otel Python 1.25/0.46b0 upgrades.

1. Re-adds `test_instrumentation` with tox environment, to run on GH before layer build
2. Removes a custom checkout that's no longer needed

See ticket comment for test layer and traces.


solarwinds_apm/version.py
```diff
@@ -1 +1 @@
-__version__ = "1.7.0"
+__version__ = "2.0.0.0"
```
jerrytfleung, line 1: Is it a typo?

tammy-baylis-swi: Nope, I do test releases with a 4th "dev" digit

PR #376, opened 2024-06-04
NH-79315 Update log warning to error if Python 3.7

tammy-baylis-swi:
Updates the startup case for logging a warning to logging an error if using Python 3.7, which neither APM Python nor OTel Python supports now.

APM Python will still continue to (try to) configure Otel and the SDK will run.

Is this too strong, or not strong enough?


cheempz (approved):
LGTM, error level seems appropriate.

PR #377, opened 2024-06-04
NH-81797 Add get_url_attrs from new or old semconv

tammy-baylis-swi:
Adds new "semconv helper" to collect url-related span attributes by order of precedence: new HTTP semconv > old HTTP semconv > None. This is to continue to support tracing mode per-request, if configured.


cheempz (approved):
LGTM!

PR #380, opened 2024-06-17
NH-82784 Fix builds

tammy-baylis-swi:
Fixes builds (testpypi and lambda) that were failing because package wasn't including new internal APM semconv.trace module.


cheempz (approved):
👍

PR #381, opened 2024-06-17
NH-82791 Remove deprecated docker-compose `version`

tammy-baylis-swi:
Removes docker-compose `version` key because now warned as obsolete (see [dc spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements)).


cheempz (approved):
LGTM

PR #387, opened 2024-07-03
NH-80452 Add OTLP logs export

tammy-baylis-swi:
Adds OTLP logs export by APM Python. Adds new dependency on Otel upstream `opentelemetry-exporter-otlp`, previously the responsibility of layer packaging or user environment when customizing traces/metrics export, because of new defaults for log export (see below list).

Enabled by default whether APM-proto or OTLP traces/metrics are exported. Enabled by default whether outside or inside lambda environment. Configurable with Otel's `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` (env var, "OTEL" below) and `SW_APM_EXPORT_LOGS_ENABLED` (env var or json config file, "SW_APM" below). OTEL has precedence over SW_APM:

| OTEL | SW_APM | Otlp log export enabled |
|---|---|---|
| unset | unset | true |
| true | unset | true |
| unset | true | true |
| true | true | true |
| false | unset | **false** |
| unset | false | false |
| false | false | false |
| false | true | **false** |
| true | false | **true** |

APM Python also sets these as defaults for OTLP logging:
1. `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL="http/protobuf"`
2. `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT="http://otel-collector:4318"`
3. `OTEL_LOGS_EXPORTER="otlp_proto_http"`

Please let me know what you think about the config options 😺


tammy-baylis-swi:
Realized that the defaults I've chosen might not be ideal if no otelcollector is running nearby. I'll address this in a separate PR.

PR #388, opened 2024-07-16
Remove support of local token bucket capacity, rate

tammy-baylis-swi:
Remove support of local token bucket capacity, rate by env var or config file. If they are set, nothing is logged because it is no longer part of the internal config.


cheempz (approved):
LGTM!

PR #393, opened 2024-07-23
NH-80452 Add OTLP logs export (v2)

tammy-baylis-swi:
Adds OTLP logs export by APM Python as opt-in feature. Adds new dependency on Otel upstream `opentelemetry-exporter-otlp`, previously the responsibility of layer packaging or user environment when customizing traces/metrics export, because of new defaults for log export (see below list).

Not enabled by default whether APM-proto or OTLP traces/metrics are exported. Not enabled by default whether outside or inside lambda environment. Configurable with Otel's `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` (env var, "OTEL" below) and `SW_APM_EXPORT_LOGS_ENABLED` (env var or json config file, "SW_APM" below). OTEL has precedence over SW_APM, i.e. the bolded results:

| OTEL | SW_APM | Otlp log export enabled |
|---|---|---|
| unset | unset | false |
| true | unset | true |
| unset | true | true |
| true | true | true |
| false | unset | false |
| unset | false | false |
| false | false | false |
| false | true | **false** |
| true | false | **true** |

APM Python distro sets these as defaults for OTLP logging for user convenience. Note: LoggingProvider and log exporters only get initialized and set if enabled as above:
1. `OTEL_EXPORTER_OTLP_LOGS_HEADERS=authorization=Bearer%20<API_TOKEN>` from service key
2. `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL=http/protobuf`
3. `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=<SWO_OTEL_COLLECTOR>/v1/logs`
4. `OTEL_LOGS_EXPORTER=otlp_proto_http`

See latest ticket comment for some manual testing results.

Please let me know if any questions or comments 😺


solarwinds_apm/distro.py
```diff
@@ -65,10 +71,42 @@ def _log_runtime(self):
         self._log_python_runtime()
         logger.info("OpenTelemetry %s/%s", sdk_version, inst_version)
 
+    def _get_token_from_service_key(self):
+        """Return token portion of service_key if set, else None"""
+        service_key = environ.get("SW_APM_SERVICE_KEY")
```
tammy-baylis-swi, line 76: This is partially duplicated from ApmConfig because Distro._configure lifetime ends before Configurator and ApmConfig are called during auto-instrumentation.

solarwinds_apm/distro.py
```diff
@@ -65,10 +71,42 @@ def _log_runtime(self):
...
+        if not service_key:
+            logger.debug("Missing service key")
```
tammy-baylis-swi, line 78: I thought I would make this and others `debug` level because the missing/poorly-formatted key "only" means the default env var for `HEADERS` is set incorrectly. It is later in auto-instrumentation during ApmConfig initialization that APM Python is disabled due to bad key.

solarwinds_apm/distro.py
```diff
@@ -65,10 +71,42 @@ def _log_runtime(self):
...
         """Configure default OTel exporter and propagators"""
         self._log_runtime()
 
+        # Set defaults for OTLP logs export by HTTP to SWO
+        header_token = self._get_token_from_service_key()
+        if not header_token:
+            logger.debug(
+                "Setting OTLP logging defaults without valid SWO token"
+            )
+        environ.setdefault(
```
tammy-baylis-swi, line 97: These 4 env vars always have defaults set. They will be overridden by user-set env vars. They will not be used by Configurator is export_logs is not enabled.

solarwinds_apm/distro.py
```diff
@@ -65,10 +71,42 @@ def _log_runtime(self):
...
+            OTEL_EXPORTER_OTLP_LOGS_HEADERS,
+            f"authorization=Bearer%20{header_token}",
+        )
+        environ.setdefault(OTEL_EXPORTER_OTLP_LOGS_PROTOCOL, "http/protobuf")
+        environ.setdefault(
+            OTEL_EXPORTER_OTLP_LOGS_ENDPOINT,
+            "https://otel.collector.na-01.cloud.solarwinds.com:443/v1/logs",
```
tammy-baylis-swi, line 104: Has to have `/v1/logs` for Otel SDK, else export will 404.

solarwinds_apm/configurator.py
```diff
@@ -437,6 +448,100 @@ def _configure_metrics_exporter(
...
+                    INTL_SWO_SUPPORT_EMAIL,
+                )
+                raise
+
+            if apm_config.is_lambda:
+                logger.debug(
+                    "Setting logs with SimpleLogRecordProcessor using %s",
+                    exporter_name,
+                )
+                logs_processor = SimpleLogRecordProcessor(exporter)
```
raphael-theriault-swi, line 528: I'm not super familiar with how Lambda instrumentation works for Python but in JS we still use batching processors and just flush at the end of the function invocation, is that something that would be possible here ?

solarwinds_apm/distro.py
raphael-theriault-swi: Are configuration and instrumentation two separate steps ?

tests/unit/test_apm_config/test_apm_config.py
```diff
@@ -673,6 +673,102 @@ def test_set_config_value_default_log_type(
...
+    def test_set_config_value_ignore_export_logs_enabled(
+        self,
+        caplog,
+        setup_caplog,
+        mock_env_vars,
+    ):
+        # Save any export_logs_enabled in os for later
+        old_expt_logs = os.environ.get("SW_APM_EXPORT_LOGS_ENABLED", None)
+        if old_expt_logs:
+            del os.environ["SW_APM_EXPORT_LOGS_ENABLED"]
```
raphael-theriault-swi, line 691: Does the Python test runner have before/after hooks that could be used to clear and restore any env vars with SW or OTEL prefixes ?

raphael-theriault-swi (commented):
Left a few questions

solarwinds_apm/distro.py
cheempz: interesting, glad you pointed this out! so maybe http/protobuf is only supported on this special endpoint, i'll ask our platform team.

solarwinds_apm/configurator.py
tammy-baylis-swi: We did try that in Python previously, since the Batch processors also have force flush functions. I think we switched to the Simple processors across the board in part because of empty metrics messages that sometimes happened: https://github.com/solarwinds/apm-python/pull/312.

This might have been fixed in the upstream since APM Python finally upgraded, via https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.21.0 So I could try switching to Batch across the board again to simplify configuration logic! Is that ok in a different ticket?

tests/unit/test_apm_config/test_apm_config.py
tammy-baylis-swi: It does, and I will work on that!

tammy-baylis-swi: Added before/after hook in [a4aaf81](https://github.com/solarwinds/apm-python/pull/393/commits/a4aaf811f9ac17315310e95a09b4ca1d831c0615) for test_apm_config and test_distro.

solarwinds_apm/configurator.py
tammy-baylis-swi: I created a new ticket! Will try switching back to Batch later.

raphael-theriault-swi (approved):
Everything I had is addressed !

PR #400, opened 2024-07-30
NH-80452 Ignore export logs config when AO collector

tammy-baylis-swi:
Ignores `SW_APM_EXPORT_LOGS_ENABLED` / `exportLogsEnabled` if collector is AO, then sets internally to `False`. Customer can still manually configure OTLP logs export with `OTEL_*` env vars that take precedence (see https://github.com/solarwinds/apm-python/pull/393).


cheempz (approved):

PR #425, opened 2024-09-03
NH-90211 Skip logs_headers setdefault in lambda

tammy-baylis-swi:
If in lambda, distro does not `setdefault` for OTLP logs headers. In regular usage, it is the otelcollector that sets this header. User or lambda wrapper can still set OTLP logs header value; there just won't be a default. I decided on this design because APM Python currently does not use `SW_APM_API_TOKEN` at all and I didn't want to introduce this dependency. This is to stop misleading warnings of `Missing service key`, which is not required for lambda.

tox tests pass on this PR. See ticket comment for manual testing.

Let me know if any questions/comments/ideas!


jerrytfleung (dismissed):
LGTM

jerrytfleung (approved):
LGTM

PR #428, opened 2024-09-05
NH-90488 lint for all supported Python versions

tammy-baylis-swi:
Update linting to run on all supported Python versions.

Lints some code to meet Python 3.12 recommendations, with `from __future__` import so Python 3.8, 3.9 still work.

 Example GH workflow run: https://github.com/solarwinds/apm-python/actions/runs/10727514035

Local command is slightly different: `make tox OPTIONS="-e lint"` is now `make tox OPTIONS="-e py312 lint"` etc


cheempz (approved):
Nice! 10 out of 10 ;)

PR #430, opened 2024-09-06
NH-90333 Add load_instrumentor condition for aws-lambda

tammy-baylis-swi:
Adds condition at custom distro's `load_instrumentor` function that is called for every eligible, installed instrumentor by upstream [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9cced9757ace5688de08ae1d58acf87c7da92cf8/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L39).

If it's the AwsLambdaInstrumentor (entry point name ["aws-lambda"](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9cced9757ace5688de08ae1d58acf87c7da92cf8/instrumentation/opentelemetry-instrumentation-aws-lambda/pyproject.toml#L37)) and we're in Lambda environment, we skip because we assume the Lambda wrapper has already loaded it.


cheempz (approved):
LGTM, thanks for the writeup and testing @tammy-baylis-swi!

PR #434, opened 2024-09-24
NH-92060 Add support for individual sqlcommenter enabling; deprecate all-or-none

tammy-baylis-swi:
Adds support for individual sqlcommenter enabling with new env var, `SW_APM_ENABLED_SQLCOMMENT`. It should be a comma-separated string of KVs paired by equals signs, where key is one of the known instrumentor names and value is a case-insensitive "boolean" string. Example: `SW_APM_ENABLED_SQLCOMMENT='django=true,sqlalchemy=false'`. False is the default, so isn't necessary to set.

Deprecates the env var `OTEL_SQLCOMMENTER_ENABLED` (documented [here](https://documentation.solarwinds.com/en/success_center/observability/content/configure/services/python/configure.htm#trace-queries)) for all-or-none enabling of sqlcommenting by instrumentors, via [this warning log](https://github.com/solarwinds/apm-python/blob/a175028cfa3ddcb695d312b23b0e70a82f63f462/solarwinds_apm/distro.py#L226-L228). There is a temporary order of precedence between both configs:

|  `OTEL_SQLCOMMENTER_ENABLED`  | `SW_APM_ENABLED_SQLCOMMENT`   |  Result  |
|----|----|---|
|  False/None  |  None  |  None enabled  |
|  False/None  |  django=true  |  Django sqlcommenting enabled   |
|  True  |  Any   |  All sqlcommenting enabled   |

For streamlining of naming, `OTEL_SQLCOMMENTER_OPTIONS` (documented [here](https://documentation.solarwinds.com/en/success_center/observability/content/configure/services/python/configure.htm#trace-queries-options)) is also deprecated in favour of new `SW_APM_OPTIONS_SQLCOMMENT`, via [this warning log](https://github.com/solarwinds/apm-python/blob/a175028cfa3ddcb695d312b23b0e70a82f63f462/solarwinds_apm/distro.py#L272-L274). They do the same thing and this is purely cosmetic. Temporary order of precedence is `OTEL_SQLCOMMENTER_OPTIONS` > `SW_APM_OPTIONS_SQLCOMMENT`.

Please let me know if any suggestions, including naming of config env vars!

### DIFF
--- a/solarwinds_apm/distro.py
+++ b/solarwinds_apm/distro.py
@@ -153,9 +153,12 @@ class SolarWindsDistro(BaseDistro):
             if SolarWindsApmConfig.calculate_is_lambda():
                 return
 
-        # Set enable for sqlcommenter. Assumes kwargs ignored if not
-        # implemented for current instrumentation library
+        # If OTEL_SQLCOMMENTER_ENABLED=true, enable for all that support it
         if self.enable_commenter():
+            # Assumes kwargs ignored if not implemented for current
+            # instrumentation library
+            # TODO: this and the `else` should set kwargs the same way
+
             # instrumentation for Flask ORM, sqlalchemy, psycopg2
             kwargs["enable_commenter"] = True
             # instrumentation for Django ORM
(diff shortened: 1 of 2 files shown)


jerrytfleung (approved):
Code looks good to me. 

But how does customer know they can configure these
```
    "django",
    "flask",
    "psycopg",
    "psycopg2",
    "sqlalchemy",
```
in `SW_APM_ENABLED_SQLCOMMENT`?

By customer doc?

tammy-baylis-swi:
> By customer doc?

Yes, I'm also working on a customer doc update for `OTEL_SQLCOMMENTER_*` deprecation and new `SW_APM_*_SQLCOMMENT` config.

PR #437, opened 2024-10-01
NH-92060 Allow case-insensitive instrumentor names when enabling sqlcomment

tammy-baylis-swi:
Noticed a small inconvenience while updating docs. This change allows case-insensitive instrumentor names when enabling sqlcomment via `SW_APM_ENABLED_SQLCOMMENT`.


cheempz (approved):
Nice catch :)

PR #450, opened 2024-11-12
NH-95827, NH-96119 Upgrade to Otel Python 1.28.1/0.49b1, testrelease 3.3.0.0

tammy-baylis-swi:
Upgrades to Otel Python 1.28.1/0.49b1. Used for testrelease to [TestPyPI as APM Python 3.3.0.0](https://test.pypi.org/project/solarwinds-apm/3.3.0.0/).

Following release of Otel Python 1.28.1/0.49b1 that removes pkg_resource uses because [deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources), also removes it on APM Python side and replaces with uses of an Otel Python util to map importlib depending on Python version (see also https://github.com/solarwinds/apm-python/pull/451). Otherwise the APM Python py312 tests fail. Also updates Makefile installations of `setuptools` (was getting new `ModuleNotFoundError` at lambda builds) and tidies deps installation calls.

Updates check_lambda_modules script because more Otel Python packages in the release follow [PEP 420](https://peps.python.org/pep-0420/).

See ticket for manual testing.


jerrytfleung (approved):
LGTM

PR #429, opened 2024-09-05
[WIP] NH-63282 Remove pkg_resources usage

tammy-baylis-swi:
WIP

`pkg_resource` is [deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources). Replace `pkg_resource` with `importlib_metadata` / `importlib.metadata` equivalents. See also ticket comment.

PR #451, opened 2024-11-12
NH-63282 Remove pkg_resources usage

tammy-baylis-swi:
pkg_resource is [deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources). Replaces pkg_resource uses with an Otel Python util to map importlib depending on Python version.

Replaces https://github.com/solarwinds/apm-python/pull/429 following release of Otel Python 1.28.1/0.49b1. Being merged into https://github.com/solarwinds/apm-python/pull/450, not main, to get tests to pass.


tammy-baylis-swi:
Merging into feature branch.

PR #470, opened 2024-12-23
NH-97751 otel.status_code, otel.status_description at APM-proto span export

tammy-baylis-swi:
Updates our custom, non-OTLP exporter to add `otel.status_code`, `otel.status_description` attributes before export, based on SDK-set span status values.


cheempz (approved):

PR #489, opened 2025-01-23
NH-100215 Update init event version lookup logs

tammy-baylis-swi:
Changes version lookup issue logs from `warning` to `debug`. My argument is that they are currently misleading when in Lambda or k8s operator, which both manually pre-install as many instrumentors as possible instead of bootstrapping only what's needed based on frameworks installed in the current environment. If there is conflict that actually prevents instrumentation, it [will be logged by upstream](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ec3c51dcd18fa747c567708d8aead21cd9081dca/opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py#L72-L76). Meanwhile the logs should still be kept if there are reporter init event issues that need deeper investigation. See ticket description for more details. Let me know what you think 😺

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -697,7 +697,7 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                         and conflict.found is None
                     ):
                         logger.debug(
-                            "Version lookup for library %s skipped due to known pydantic/tortoiseorm bootstrap conflice: %s",
+                            "Version lookup for library %s skipped due to known pydantic/tortoiseorm bootstrap conflict: %s",
                             entry_point.name,
                             conflict,
                         )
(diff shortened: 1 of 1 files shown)


cheempz (approved):
I agree dialing down the verbosity makes sense, it's not critical that we gather the instrumented library version (which IIRC is what we're after, which would be an arbitrary version the customer might install in their k8s pod or lambda). The version of the instrumentor itself i believe we take from the instrumentation scope and attach as an attribute to the span.

cheempz:
Actually, re-reading the details you put in Jira, it seems we are successfully getting the version of the instrumented component if it is actually installed. So no loss of functionality and maybe what we're running into is how for k8s and perhaps lambda, the fact that instrumentors are installed (regardless their target existing in the environment) surfaces a "maybe expected" exception in the dependency checking.

tammy-baylis-swi:
> maybe what we're running into is how for k8s and perhaps lambda, the fact that instrumentors are installed (regardless their target existing in the environment) surfaces a "maybe expected" exception in the dependency checking.

Yes I think so!

PR #494, opened 2025-01-29
NH-100646 Default OTLP auth header always in lambda

tammy-baylis-swi:
Simplifies APM Python custom distro by removing one unnecessary `is_lambda` check.

Change is that now, in lambda environment, default values for `OTEL_EXPORTER_OTLP_(TRACES|METRICS|LOGS)_HEADERS` always get set to `f"authorization=Bearer%20{header_token}"` even if `header_token` is empty, which should normally be the case. This does not affect export to otel-collector layer on localhost -- see ticket comment for more test info. This is again for default values, so user can always override with their own `HEADERS` set in environment.

Works the same as before outside of lambda.

Also adds more unit tests and clarifies existing ones.

### DIFF
--- a/tests/unit/test_distro.py
+++ b/tests/unit/test_distro.py
@@ -401,8 +401,9 @@ class TestDistro:
         )
         distro.SolarWindsDistro()._configure()
         assert os.environ.get(OTEL_EXPORTER_OTLP_TRACES_HEADERS) is None
-        assert os.environ.get(OTEL_EXPORTER_OTLP_METRICS_HEADERS) is None
-        assert os.environ.get(OTEL_EXPORTER_OTLP_LOGS_HEADERS) is None
+        # Still get set for metrics and logs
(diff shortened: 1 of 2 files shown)


tests/unit/test_distro.py
```diff
@@ -347,6 +352,18 @@ def test__configure_traces_export_env_defaults_valid_protocol_grpc(self, mocker)
         assert os.environ.get(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT) == "https://otel.collector.na-01.cloud.solarwinds.com:443/v1/traces"
         assert os.environ.get(OTEL_EXPORTER_OTLP_TRACES_HEADERS) == "authorization=Bearer%20foo-token"
 
+    def test_configure_set_otlp_header_defaults_not_lambda_no_header_token(self, mocker):
```
tammy-baylis-swi, line 355: This test coverage was missing before

cheempz (approved):
LGTM, thanks for the writeup and tests @tammy-baylis-swi!

PR #496, opened 2025-01-29
NH-100646 Internal service_key updates more general

tammy-baylis-swi:
Changes APM-internal service_key updates to be more general. Removes is_lambda check so that `_update_service_key_name` is always called. Updates that function to avoid `IndexError` for cases like when in Lambda and `SW_APM_SERVICE_KEY` is unnecessarily set as a weird value. Adds missing test coverage.


cheempz (approved):
LGTM :)

PR #497, opened 2025-01-29
NH-98561 Add ApmConfig.calculate_is_legacy

tammy-baylis-swi:
Note: this will merge into epic feature branch `NH-79025-otlp-by-default`.

Adds first parts for APM Python to support `SW_APM_LEGACY` by reading from environment variable (highest precedence) or `legacy` in JSON config file, else default `False`. APM Python does not use this for anything yet.

Why are `calculate_is_legacy` and now `get_cnf_dict` classmethods in ApmConfig? Because:
1. Distro needs them to change its `setdefault` calls in a future PR
2. Distro and Configurator can't share one ApmConfig due to lifecycles when [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) is used for auto-instrumentation. So the config file does get read twice now at instrumented service startup.

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -246,6 +246,20 @@ class SolarWindsApmConfig:
         )
 
     @classmethod
+    def calculate_is_legacy(cls) -> bool:
+        """Checks if agent is running in a legacy environment.
+        Order of precedence: Environment Variable > config file > default False
+        """
+        is_legacy = False
+        cnf_dict = cls.get_cnf_dict()
+        if cnf_dict:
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM! I guess if `get_cnf_dict` would be called multiple times, the first call could cache it--but minor and may not be worth it given we'd likely revamp as part of future adoption of upstream declarative config anyway.

PR #498, opened 2025-01-30
NH-98561 Add ApmConfig support for SW_APM_LEGACY

tammy-baylis-swi:
Note: this will merge into epic feature branch `NH-79025-otlp-by-default`.

Adds ApmConfig support for reading `SW_APM_LEGACY` env var and `legacy` in JSON config file. Default is `False`. Just stores the value internally for now.

This is needed for future PRs to e.g. decide when to init components like Reporter and InboundMetricsSpanProcessor. This is similar to https://github.com/solarwinds/apm-python/pull/497 but to make the value available internally to the Configurator and the rest of APM's life cycle (see also [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) used for auto-instrumentation).

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -113,6 +113,7 @@ class SolarWindsApmConfig:
             "transaction_name": None,
             "export_logs_enabled": False,
             "export_metrics_enabled": False,
+            "legacy": False,
         }
         self.is_lambda = self.calculate_is_lambda()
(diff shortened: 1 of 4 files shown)


cheempz (approved):
LGTM!

PR #503, opened 2025-02-01
NH-100646 _get_extension now based on if Legacy, not if Lambda

tammy-baylis-swi:
Note: This will merge into epic feature branch `NH-79205-otlp-by-default`

`_get_extension_components` is now based on `SW_APM_LEGACY` false/true, instead of "is lambda". I'm keeping two separate try-except for now around import of `solarwinds_apm.extension.oboe` then import of OboeAPI and options, in case of upcoming updated extension builds. Also updates `str(ApmConfig)` debug helper to include OboeAPI.


cheempz (approved):
LGTM!

PR #504, opened 2025-02-01
NH-100646 Create init event if Legacy, instead of if not-Lambda

tammy-baylis-swi:


solarwinds_apm/configurator.py
```diff
@@ -808,10 +808,6 @@ def _create_init_event(
         keys: dict = None,
     ) -> Any:
         """Create a Reporter init event if the reporter is ready."""
-        if apm_config.is_lambda:
-            logger.debug("Skipping init event in lambda")
-            return None
-
```
tammy-baylis-swi, line 814: Removed this because `_configure` is the only caller of this `_create_init_event`, and the former checks if legacy already

tests/unit/test_configurator/test_configurator_create_init_event.py
```diff
@@ -9,35 +9,6 @@
...
-        mock_extension,
-        mock_apmconfig_enabled_is_lambda,
-    ):
-        trace_mocks = get_trace_mocks(mocker)
-
-        test_configurator = configurator.SolarWindsConfigurator()
-        result = test_configurator._create_init_event(
-            mock_extension.Reporter,
-            mock_apmconfig_enabled_is_lambda,
-        )
```
tammy-baylis-swi, line 27: Removed this test because function doesn't do this check anymore (see [comment](https://github.com/solarwinds/apm-python/pull/504/files#r1939864361)).

raphael-theriault-swi (approved):

PR #507, opened 2025-02-04
NH-98561 Update configure_otel_components conditions

tammy-baylis-swi:
Note: This will merge into epic feature branch `NH-79205-otlp-by-default`

This makes initialization of custom OTel components explicitly dependent on Legacy Mode (`SW_APM_LEGACY`) within `_configure_otel_components`. The implicit dependency on supported env vars like `OTEL_METRICS_EXPORTER` still exists in the helper functions to keep supporting upstream config options. But now there is a clear line at 'top level' for internal component setup. Traces exporter is always configured with either default values set by SolarWindsDistro or user-set `OTEL_TRACES_EXPORTER`.

This also centralizes agent enabled vs disabled behaviour into `_configure_otel_components`, rather than having some but not all of the helpers doing it (probably some code smell).

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -159,6 +159,8 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                     apm_config,
                     oboe_api,
                 )
+
+                self._configure_logs_exporter(apm_config)
             else:
                 # Export APM metrics by APM-proto
                 self._configure_inbound_metrics_span_processor(
@@ -178,7 +180,6 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                 apm_fwkv_manager,
                 apm_config,
             )
-            self._configure_logs_exporter(apm_config)
 
         else:
             # Warning: This may still set OTEL_PROPAGATORS if set because OTel API defaults
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -151,29 +151,25 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                 apm_txname_manager,
             )
 
-            if apm_config.get("legacy") is False:
-                # Export APM metrics by OTLP
+            if apm_config.get("legacy") is True:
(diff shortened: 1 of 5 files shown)


tests/unit/test_configurator/test_configurator_metrics_exporter.py
```diff
@@ -16,26 +16,6 @@
 
 
 class TestConfiguratorMetricsExporter:
-    def test_configure_metrics_exporter_disabled(
```
tammy-baylis-swi, line 19: Removed this and other similar test because this helper now assumes top `_configure` function does the enabled/disabled check.

solarwinds_apm/configurator.py
```diff
@@ -132,62 +132,70 @@ def _configure_otel_components(
...
         self,
         apm_config: SolarWindsApmConfig,
         reporter: "Reporter",
         oboe_api: "OboeAPI",
     ) -> None:
-        """Always configure SolarWinds OTel sampler, or none if disabled"""
-        if not apm_config.agent_enabled:
-            logger.error("Tracing disabled. Using OTel no-op tracer provider.")
-            trace.set_tracer_provider(trace.NoOpTracerProvider())
-            return
```
tammy-baylis-swi, line 190: Removed here and in other helpers because calling `_configure_otel_components` does one enabled/disabled check.

solarwinds_apm/configurator.py
```diff
@@ -396,15 +397,9 @@ def _configure_metrics_exporter(
...
+        """Configure OTel OTLP metrics exporters if enabled.
         Settings precedence: OTEL_* > SW_APM_EXPORT_METRICS_ENABLED.
         Links to new metric readers and global MeterProvider."""
-        if not apm_config.agent_enabled:
-            logger.error(
-                "APM Python library disabled. Cannot set metrics exporters."
-            )
-            return
-
         if not apm_config.get("export_metrics_enabled"):
```
tammy-baylis-swi, line 403: This will be updated in another PR

cheempz (approved):
LGTM!

PR #499, opened 2025-01-31
NH-98561 Update OTLP setdefaults with legacy check and cleanup

tammy-baylis-swi:
Note: This will merge into epic feature branch `NH-79205-otlp-by-default`.

This updates APM Python's `Distro` component, which is called first when [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) is used for auto-instrumentation, to more clearly prepare for OTLP-by-default. I'm changing the parts used to `setdefault` of exporter configurations. All values can still be configured (i.e. override defaults) by user via `OTEL_*` env vars. It is the `Configurator` component's responsibility to load and init the exporters later according to final config.

Key code changes:

(1)
For users, it used to be that setting an invalid protocol like `OTEL_EXPORTER_OTLP_PROTOCOL: "foo"` (not `http/protobuf` nor `grpc`) would be ignored and a safeguard defaults would be set: APM-proto exporter for traces + OTLP exporter for metrics and logs (a confusing mix!). This is not advertised. Now, if invalid protocol, all _default_ exporters will be `None` and a warning is logged. See also [this test case](https://github.com/solarwinds/apm-python/blob/dc3c7404579afd414e455fcb8c05d88b98ba6a57/tests/unit/test_distro.py#L475-L498).

(2)
For users, it used to be that invalid protocol OR not setting protocol at all would setdefault `OTEL_TRACES_EXPORTER` as APM-proto (also confusing!) Now, the new `calculate_is_legacy` method is used. If legacy, set APM-proto as default. If not, set OTLP by http/proto as default. However, if an invalid protocol was provided, exporter will be `None` -- see (1) above. See also [this test case](https://github.com/solarwinds/apm-python/blob/dc3c7404579afd414e455fcb8c05d88b98ba6a57/tests/unit/test_distro.py#L716-L741) vs [this other test case](https://github.com/solarwinds/apm-python/blob/dc3c7404579afd414e455fcb8c05d88b98ba6a57/tests/unit/test_distro.py#L743-L771).

(3)
Refactors and renames `_configure_(traces|metrics|logs)_export_env_defaults` to `_configure_(traces|metrics|logs)_export_otlp_env_defaults` to cut duplicate code and imports.

(4)
[...]

PR #508, opened 2025-02-04
NH-98561 Breaking: Update metrics/logs-from-OTel opt-in and existing config

tammy-baylis-swi:

### Before this PR

Before these changes, `SW_APM_EXPORT_(METRICS|LOGS)_ENABLED` was a convenience config for enabling/disabling all metrics and logs export by OTLP unless AO collector set. But now we can and will support both an AO collector and an OTLP collector.

### After this PR - main changes

This PR changes the role of `SW_APM_EXPORT_(METRICS|LOGS)_ENABLED` (breaking!) to support these cases:
	
If `SW_APM_LEGACY=true` and `SW_APM_EXPORT_(METRICS|LOGS)_ENABLED=true`, then library still exports SWO metrics by APM-proto. In addition, Otel Python instrumentor metrics and logs will export by OTLP. If `ENABLED=false` (default), then only do SWO metrics by APM-proto.
	
If not legacy and `SW_APM_EXPORT_(METRICS|LOGS)_ENABLED=false`, then library still exports SWO metrics and logs by OTLP. But Otel Python instrumentor metrics and logs won't be exported (if implemented) by passing NoOp providers at `load_instrumentor`. If `ENABLED=true` (default for metrics only), both sets are exported.

Note: `SW_APM_EXPORT_METRICS_ENABLED` is now True by default. `SW_APM_EXPORT_LOGS_ENABLED` is still False by default.

### Some more info

The Distro class now caches `_cnf_dict` as contents read from a valid JSON config file once at singleton creation. This is so the distro only reads the file once to calculate (1) if legacy during `_configure`, (2) if Otel Python metrics/log enabled for each call to `load_instrumentor`. Auto-instrumentation still reads the file twice (another time by the Configurator to get all other settings); Distro and Configurator can't share one ApmConfig due to lifecycles when [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) is used.

All configs like `OTEL_METRICS_EXPORTER`, `PROTOCOL`, `ENDPOINT`, etc are still supported.

APM Python still respects `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` if configured by user as before. This behaviour will be updated later (added todos).

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -4,6 +4,8 @@
 #
 # Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
 
+# pylint: disable=too-many-lines
+
 import json
 import logging
 import os
@@ -169,10 +171,6 @@ class SolarWindsApmConfig:
         # (Re-)Calculate config if AppOptics
         self.metric_format = self._calculate_metric_format()
         self.certificates = self._calculate_certificates()
-        self.__config["export_logs_enabled"] = self._calculate_logs_enabled()
-        self.__config["export_metrics_enabled"] = (
-            self._calculate_metrics_enabled()
-        )
 
         logger.debug("Set ApmConfig as: %s", self)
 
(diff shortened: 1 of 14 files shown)


tests/unit/test_apm_config/test_apm_config.py
```diff
@@ -539,104 +534,6 @@ def test_calculate_certificates_ao_prod_trustedpath_file_present(self, mocker):
         mock_get_public_cert.configure_mock(return_value="foo")
         assert apm_config.SolarWindsApmConfig()._calculate_certificates() == "bar"
 
-    def test_calculate_logs_enabled_no_collector_enabled(self, mocker):
```
tammy-baylis-swi, line 542: Removing this test coverage because calculating metrics/logs "enabled" no longer depends on collector.

tests/unit/test_configurator/test_configurator_logs_exporter.py
```diff
@@ -115,61 +115,7 @@ def test_configure_logs_exporter_otel_none_sw_false(
         mock_blprocessor.assert_not_called()
         mock_logginghandler.assert_not_called()
 
-    def test_configure_logs_exporter_otel_true_sw_false(
```
tammy-baylis-swi, line 118: Removing this test coverage because logs/metrics exporters are always configured now based on `OTEL_(LOGS|METRICS)_EXPORTER` etc, with defaults set by Distro (e.g. [this PR](https://github.com/solarwinds/apm-python/pull/499/files)).

solarwinds_apm/apm_config.py
```diff
@@ -272,6 +276,66 @@ def calculate_is_lambda(cls) -> bool:
...
+        """Return if export of instrumentor logs telemetry enabled.
+        Invalid boolean values are ignored.
+        Order of precedence: Environment Variable > config file > default.
+        Default is True is not legacy, False if legacy.
+        Optional cnf_dict is presumably already from a config file, else a call
+        to get_cnf_dict() is made for a fresh read."""
+        if is_legacy:
+            logs_enabled = False
+        else:
+            logs_enabled = True
```
raphael-theriault-swi, line 294: I vaguely remember having a conversation about not turning on logs export by default based on pricing uncertainty (as in to avoid customers getting a surprise bill because they're now sending in a bunch of logs), has that been settled ?

tammy-baylis-swi: Ah good catch. No, I did not come to a conclusion on this. I will check and get back!

cheempz: That's right, we don't want to enable logs export by default, it should be explicitly opted into by the customer.

tammy-baylis-swi: Thank you both! Updated in [34c8513](https://github.com/solarwinds/apm-python/pull/508/commits/34c8513eaab62ac82744ffe32e5728b167c6c60f), and PR description too.

solarwinds_apm/configurator.py
```diff
@@ -157,6 +157,18 @@ def _configure_otel_components(
                     apm_txname_manager,
                     apm_config,
                 )
+                # While in legacy mode, user can also opt into exporting
+                # Otel instrumentor metrics and logs by OTLP.
+                # Default values are set by SolarWindsDistro; user can customize.
+                if apm_config.get("export_metrics_enabled") is True:
+                    self._configure_metrics_exporter(apm_config)
+                    self._configure_otlp_metrics_span_processors(
```
cheempz, line 165: In legacy mode i assume the response time metric (`trace.service.response_time`) is being calculated by our `SolarWindsInboundMetricsSpanProcessor` and exported via apm-proto. In this case, do we need to `_configure_otlp_metrics_span_processors` since that is the OTLP counterpart to processing spans for response time?

tammy-baylis-swi: Yes, you're right! Removed in [27204c4](https://github.com/solarwinds/apm-python/pull/508/commits/27204c4c224c7aefb1092cbcf7ffa21b97d5ffa8) and updated in-code comments. Also saw an opportunity to simplify the helper in [20eb231](https://github.com/solarwinds/apm-python/pull/508/commits/20eb2319c78e3a4c81a74f158a433e531c241b7d) as exporters always init first anyway for the non-legacy case.

cheempz (approved):
LGTM!

PR #509, opened 2025-02-06
NH-98561 Remove unnecessary Lambda-specific defaults

tammy-baylis-swi:

Removes default configs set by the wrapper for APM Python Lambda layers because no longer necessary, as these are now the defaults outside Lambda too (see https://github.com/solarwinds/apm-python/pull/499, https://github.com/solarwinds/apm-python/pull/508).


raphael-theriault-swi (approved):
🎉

PR #510, opened 2025-02-06
NH-98561 Update integration tests

tammy-baylis-swi:

Updates the APM Python integration/functional tests so that OTLP metrics/logs generation and export are checked when enabled. Overall changes:

1. Moves the original tests to `legacy` subdir, and they're the same besides some var renames.
2. Adds new tests to `otlp` subdir. They test the same scenarios as `legacy` but with different liboboe mocks and added OTLP metrics, logs assertions.
3. Factors out more general `TestBaseSw` class (with a much shorter name), extended by `TestBaseSwLegacy` and `TestBaseSwOtlp` for their respective tests.

As previously, these tests do partial, manual loading of APM Python otel components so that traces/metrics/logs can be captured by the `InMemory*` test utils and checked there, instead of actual export to backend/file/etc.

Please let me know if any questions/comments!


tests/integration/otlp/test_base_sw_otlp.py
```diff
@@ -0,0 +1,219 @@
...
+                "network.protocol.version": "1.1",
+                "http.response.status_code": 200,
+                "http.route": "/test_trace/"
+            },
+        )
+
+    def assertRequestsMetrics(self, status_code=True):
+        """Assert client metrics from test app Requests instrumentation.
+        Instrumentor sometimes does not include status_code attribute,
+        i.e. decision is not-sampled."""
```
cheempz, line 138: minor aside re: status_code only available if span is sampled, seems a bug in upstream.

tammy-baylis-swi: !!!! It might be this!! https://github.com/open-telemetry/opentelemetry.io/issues/6274

cheempz (approved):
LGTM, pretty neat that OTel python provides the in-memory exporter/readers :)

PR #519, opened 2025-02-13
NH-98561 Update logging auto-instrumentation support

tammy-baylis-swi:
Note: this will merge into feature branch `NH-79205-otlp-by-default`.

Updates APM Python support of `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED`:

1. `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` is False by default; it's an opt-in for users.
2. If True, then logs SDK handler is created and added.
3. Logging provider and exporters are always set up, with latter depending on `OTEL_LOGS_EXPORTER` (otlp http by default).

### DIFF
--- a/tests/unit/test_configurator/conftest.py
+++ b/tests/unit/test_configurator/conftest.py
@@ -6,6 +6,8 @@
 
 import pytest
 
+from solarwinds_apm.apm_config import SolarWindsApmConfig
+
 # ==================================================================
 # Configurator stdlib fixtures
 # ==================================================================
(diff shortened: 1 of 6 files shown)


tests/unit/test_configurator/conftest.py
```diff
@@ -196,6 +198,7 @@ def get_side_effect(param):
             "is_lambda": is_lambda,
             "extension": mock_ext,
             "oboe_api": mocker.Mock(),
+            "convert_to_bool": SolarWindsApmConfig.convert_to_bool,
```
tammy-baylis-swi, line 201: I am being lazy and keeping the elaborate test mocks but using the real `convert_to_bool` function to mock this function.

raphael-theriault-swi (approved):

PR #523, opened 2025-02-14
NH-102113 Add ClickInstrumentor to image build

tammy-baylis-swi:
Add ClickInstrumentor to image build now that we've upgraded Otel Python.


cheempz (approved):
👍

PR #524, opened 2025-02-14
NH-102114 Add support for SW_APM_ENABLED_SQLCOMMENT_ATTRIBUTE

tammy-baylis-swi:
Add support for new config `SW_APM_ENABLED_SQLCOMMENT_ATTRIBUTE`. It is "true" by default for APM Python for all instrumentors. It can be set to "false" for each to opt out. When true _and_ `SW_APM_ENABLED_SQLCOMMENT` is true for that instrumentor, then it will add sqlcomment to `db.statement` attribute.

It accepts a string representing a list of instrumentor keys and their values, delimited by commas. Accepted instrumentor names are: sqlalchemy, django, flask, psycopg2, and psycopg2. Other keys will be ignored.

Example:
```
export SW_APM_ENABLED_SQLCOMMENT_ATTRIBUTE="psycopg2=true,sqlalchemy=false"
```

### DIFF
--- a/solarwinds_apm/distro.py
+++ b/solarwinds_apm/distro.py
@@ -226,7 +226,9 @@ class SolarWindsDistro(BaseDistro):
         individual instrumentor by upstream sitecustomize.
 
         For enabling sqlcommenting, SW_APM_ENABLED_SQLCOMMENT enables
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM!

PR #528, opened 2025-02-18
NH-102110 Update auto-instrumentation image APM dep

tammy-baylis-swi:
Bump auto-instrumentation image APM Python dependency for next publish.


cheempz (approved):
👍

PR #530, opened 2025-02-19
NH-102374 Docker scout scans consistency

tammy-baylis-swi:
Docker scout scan results more consistent like APM Java. Splits image build with first `load=true`, and removes fs scan. See ticket comment for result.


tammy-baylis-swi:
Going to re-open later with additional changes.

PR #531, opened 2025-02-19
NH-102271 Add image APM version bump to 'Create Release PR' action

tammy-baylis-swi:
Updates the "Create Release PR" workflow to also bump the APM Python version in [image/requirements-nodeps.txt](https://github.com/solarwinds/apm-python/blob/main/image/requirements-nodeps.txt#L1). This is so "Publish APM Python Auto-Instrumentation" can be run, manually and minutes after PyPI updates are picked up by GH, after PyPI publish is complete but with less to remember.

This also fixes the existing [version.py](https://github.com/solarwinds/apm-python/blob/main/solarwinds_apm/version.py#L1) update logic that was unnecessarily adding newlines.

Example test PR: https://github.com/solarwinds/apm-python/pull/533


cheempz (approved):
LGTM, thanks for that example PR :)

PR #535, opened 2025-02-20
NH-102374 update scout scans

tammy-baylis-swi:
Replaces https://github.com/solarwinds/apm-python/pull/530

Docker scout scan results more consistent like APM Java. Splits image build with first load=true, and removes fs scan. Adds upload of sarif results. See ticket comment for result.

### DIFF
--- a/.github/workflows/build_publish_image_autoinstrumentation.yaml
+++ b/.github/workflows/build_publish_image_autoinstrumentation.yaml
@@ -58,12 +58,23 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push
+      - name: Build and push - amd64
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          load: true
+          context: image
+          platforms: linux/amd64
+          build-args: version=${{ env.VERSION }}
+          tags: ${{ steps.meta.outputs.tags }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_publish_image_autoinstrumentation.yaml
```diff
@@ -7,12 +7,16 @@
 name: "Publish APM Python Auto-Instrumentation"
 
 on:
+  push:
+    branches:
+      - NH-102374-update-scout-scans
```
michael-sabrnak-swi, line 12: Just curious - do you want to change this from `- NH-102374-update-scout-scans` to `pull_request`? So it gets triggered on every PR.

But looks good anyway thanks for implementing SCOUT with GHAS!

michael-sabrnak-swi (dismissed):

.github/workflows/build_publish_image_autoinstrumentation.yaml
tammy-baylis-swi: Ah, I forgot to remove my test trigger -- thanks! Removed in [4cc5164](https://github.com/solarwinds/apm-python/pull/535/commits/4cc5164056c7ad827621cfed7e0b32d777b5e13a)

I might add `pull_request` another way in another PR. The builds are very slow at the moment.

cleverchuk (approved):
LGTM

PR #537, opened 2025-03-01
NH-103097 Remove image build "push" condition for GH logins

tammy-baylis-swi:
Remove "push" condition for GH logins during image build, which is currently causing `401 Unauthorized`. [Example failed workflow](https://github.com/solarwinds/apm-python/actions/runs/13599288367/job/38022561971). Currently we run this through manual trigger, not on-push.


cheempz (approved):
👍

Issue #538, opened 2025-03-03
Breaking celery beat when using sqs


twiggy:
Using against celery beat v5.4.0. Stack trace shows problem in propagator.py. Using sqs queue for celery. boto3 is version 1.37.1.  Maybe broken for other transports as well, but see boto3sqs OTel in the trace.

`File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/celery/beat.py", line 404, in apply_async
    return task.apply_async(entry_args, entry_kwargs,
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/celery/app/task.py", line 575, in apply_async
    return app.send_task(
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/celery/app/base.py", line 788, in send_task
    amqp.send_task_message(P, name, message, **options)
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/celery/app/amqp.py", line 510, in send_task_message
    ret = producer.publish(
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/kombu/messaging.py", line 186, in publish
    return _publish(
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/kombu/connection.py", line 556, in _ensured
    return fun(*args, **kwargs)
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/kombu/messaging.py", line 208, in _publish
    return channel.basic_publish(
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/kombu/transport/virtual/base.py", line 614, in basic_publish
    return self._put(routing_key, message, **kwargs)
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/kombu/transport/SQS.py", line 477, in _put
    c.send_message(**kwargs)
  File "/home/deployer/.pyenv/versions/3.10.7/envs/celeryman_env/lib/python3.10/site-packages/opentelemetry/instrumentation/boto3sqs/__init__.py", line 237, in send_wrapper
[...]

PR #539, opened 2025-03-07
NH-103629 Propagator custom `inject` checks for when baggage, tracestate set as dict

tammy-baylis-swi:
Addresses https://github.com/solarwinds/apm-python/issues/538

Updates APM Python custom propagator to handle `baggage` and `tracestate` if injected by some upstream OTel instrumentors as dict instead of str:

* [opentelemetry-instrumentation-boto3sqs](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d5dce5de996efcf2c2f85eed787c13f831095377/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L77-L89)
* [opentelemetry-instrumentation-botocore](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d5dce5de996efcf2c2f85eed787c13f831095377/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/_messaging.py#L26-L31)

APM-sanitized, non-empty baggage is still always re-injected as string.

### DIFF
--- a/solarwinds_apm/propagator.py
+++ b/solarwinds_apm/propagator.py
@@ -136,8 +136,10 @@ class SolarWindsPropagator(textmap.TextMapPropagator):
 
         # Remove any baggage stored for custom transaction naming
         if baggage_header:
-            # Note: OTel instrumentors of messaging systems (e.g. boto3) may set
+            # Note: OTel instrumentors of messaging systems (e.g. boto3) may inject
             # baggage header value as a dictionary, not a string.
+            # APM Python SDK always sets baggage header as a string because the same
+            # instrumentors' getters extract with expectation of string.
             if isinstance(baggage_header, dict):
                 baggage_stringvalue = baggage_header.get("StringValue")
                 handle_baggage_header(baggage_stringvalue)
--- a/solarwinds_apm/propagator.py
+++ b/solarwinds_apm/propagator.py
@@ -125,7 +125,7 @@ class SolarWindsPropagator(textmap.TextMapPropagator):
             carrier, self._TRACESTATE_HEADER_NAME, trace_state.to_header()
         )
 
-        def handle_baggage_header(baggage_header):
+        def _handle_baggage_header(baggage_header):
             sanitized_baggage = self.remove_custom_naming_baggage_header(
                 baggage_header
             )
@@ -144,9 +144,9 @@ class SolarWindsPropagator(textmap.TextMapPropagator):
             # inject header values as dictionary, not a string.
             if isinstance(baggage_header, dict):
                 baggage_stringvalue = baggage_header.get("StringValue")
-                handle_baggage_header(baggage_stringvalue)
+                _handle_baggage_header(baggage_stringvalue)
             elif isinstance(baggage_header, str):
-                handle_baggage_header(baggage_header)
+                _handle_baggage_header(baggage_header)
             # Else: leave baggage header as is in carrier
 
     def remove_custom_naming_baggage_header(
--- a/solarwinds_apm/propagator.py
+++ b/solarwinds_apm/propagator.py
@@ -83,6 +83,10 @@ class SolarWindsPropagator(textmap.TextMapPropagator):
 
         # Prepare carrier with carrier's or new tracestate
         trace_state = None
+        # Note: OTel Propagation API callers (OTel instrumentors) may
+        # inject header values as dictionary, not a string.
+        if isinstance(trace_state_header, dict):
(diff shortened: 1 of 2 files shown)


cheempz (commented):
Ugh, i took a while trying to understand what's going on... afaict [Boto3SQSGetter](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d5dce5de996efcf2c2f85eed787c13f831095377/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L61-L74) does expect a map instead of string  

> if not isinstance(msg_attr, Mapping)

and i'm also not sure if the error [twiggy saw](https://github.com/solarwinds/apm-python/issues/538#issuecomment-2695773767) with 

>   File "solarwinds_apm/propagator.py", line 96, in inject
              trace_state = TraceState.from_header([trace_state_header])

will be addressed by current change.

If i'm understanding things correctly, this is happening because our solarwinds_propagator is part of the composite propagator that "kicks in" whenever an instrumentor calls propagator [extract](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d5dce5de996efcf2c2f85eed787c13f831095377/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L201) or [inject](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d5dce5de996efcf2c2f85eed787c13f831095377/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L237) -- which in this case the getter/setter is paired (used against) the SQS Message Attribute object as carrier, not the TextMap carrier/HTTP header that our propagator assumes it is working against.

If this is true, seems we'd need to ensure our propagator plays well extracting/injecting into different carrier types. Side curiosity is, since we specify the input carrier type, why doesn't that seem to be honoured? Feels like the SDK should only apply a propagator in the composite chain if the carrier matches?

Another thought is our custom `extract` logic is not applicable to SQS / messaging systems since they all concern trigger tracing, maybe it can no-op unless the right carrier type. Our custom `inject` logic is applicable though, since we'd want sampling on the consumer side to take into account the `tracestate: sw=`... may need to ensure the trace_state_header handling also can handle dictionary.

cheempz:
> Side curiosity is, since we specify the input carrier type, why doesn't that seem to be honoured? Feels like the SDK should only apply a propagator in the composite chain if the carrier matches?

Scratch that, i was ignorant and after looking at [propagator.textmap](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/propagators/textmap.py) it seems the carrier and carrier value types are generic by design.

solarwinds_apm/propagator.py
```diff
@@ -121,22 +121,32 @@ def inject(
             carrier, self._TRACESTATE_HEADER_NAME, trace_state.to_header()
         )
 
-        # Remove any baggage stored for custom transaction naming
-        if baggage_header:
-            baggage_header = self.remove_custom_naming_baggage_header(
-                baggage_header,
+        def handle_baggage_header(baggage_header):
```
jerrytfleung, line 124: Do we want to add a leading underscore to indicate it is private to the function?

jerrytfleung (approved):
It looks good to me. Just left a `nits` comment

solarwinds_apm/propagator.py
tammy-baylis-swi: Addressed in [a41b577](https://github.com/solarwinds/apm-python/pull/539/commits/a41b5771c54ab2d7c22733247adc5554b8a1d1b8)

tammy-baylis-swi:
Regarding
> File "solarwinds_apm/propagator.py", line 96, in inject
trace_state = TraceState.from_header([trace_state_header])

Addressed in [46bf0e2](https://github.com/solarwinds/apm-python/pull/539/commits/46bf0e2dcd38f61e1925061f55a3729b40004487) for similar multi-type handling.

cheempz (approved):
LGTM, thanks for the revisit @tammy-baylis-swi! As discussed this is a targeted fix for the way SQS instrumentation propagates context, and longer term the ideal solution is a general one where our customization either does not have to access the data in the carrier at all, or only access it through the supplied getter/setter.

PR #551, opened 2025-03-24
NH-105445 Breaking: update OTLP metrics export defaults and configs

tammy-baylis-swi:
Replaces https://github.com/solarwinds/apm-python/pull/508

Breaking changes:

1. APM Python will now always generate and export SWO/vendor OTLP metrics by default (previously an opt-in), like other APM libraries.
2. Behaviour of `SW_APM_EXPORT_METRICS_ENABLED` is changed. It is now a config to opt out of non-SWO metrics, e.g. from Otel Python instrumentors.

I will put in a documentation update PR separately.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -396,21 +396,8 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
         self,
         apm_config: SolarWindsApmConfig,
     ) -> None:
-        """Configure OTel OTLP metrics exporters if enabled and agent enabled.
-        Settings precedence: OTEL_* > SW_APM_EXPORT_METRICS_ENABLED.
+        """Configure OTel OTLP metrics exporters.
         Links to new metric readers and global MeterProvider."""
-        if not apm_config.agent_enabled:
-            logger.error(
-                "APM Python library disabled. Cannot set metrics exporters."
-            )
-            return
-
-        if not apm_config.get("export_metrics_enabled"):
-            logger.debug(
-                "APM OTLP metrics export disabled. Skipping init of metrics exporters"
-            )
-            return
-
         # SolarWindsDistro._configure does setdefault before this is called
         environ_exporter = os.environ.get(
             OTEL_METRICS_EXPORTER,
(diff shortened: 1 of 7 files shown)


solarwinds_apm/configurator.py
```diff
@@ -396,21 +396,8 @@ def _configure_metrics_exporter(
...
     ) -> None:
-        """Configure OTel OTLP metrics exporters if enabled and agent enabled.
-        Settings precedence: OTEL_* > SW_APM_EXPORT_METRICS_ENABLED.
+        """Configure OTel OTLP metrics exporters.
         Links to new metric readers and global MeterProvider."""
-        if not apm_config.agent_enabled:
-            logger.error(
-                "APM Python library disabled. Cannot set metrics exporters."
-            )
-            return
```
tammy-baylis-swi, line 406: This check is already made by the caller, so removing from here.

solarwinds_apm/distro.py
```diff
@@ -67,6 +68,21 @@
 class SolarWindsDistro(BaseDistro):
     """OpenTelemetry Distro for SolarWinds reporting environment"""
 
+    _cnf_dict = None
+    _instrumentor_metrics_enabled = None
+
+    def __new__(cls, *args, **kwargs):
+        # Maintain singleton pattern and cache SW APM user config
+        # for Distro lifecycle at auto-instrumentation
```
tammy-baylis-swi, line 76: Implementation detail: the Distro class now caches `_cnf_dict` as contents read from a valid JSON config file once at singleton creation. This is so the distro only reads the file once to calculate if Otel Python metrics enabled for each call to `load_instrumentor`. Auto-instrumentation still reads the file twice (another time by the Configurator to get all other settings); Distro and Configurator can't share one ApmConfig due to lifecycles when [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) is used.

PR #557, opened 2025-03-26
WIP/PoC: ServiceEntrySpanProcessor caches span for set_transaction_name

tammy-baylis-swi:
WIP/PoC: `ServiceEntrySpanProcessor` (formerly ServiceEntry**Id**SpanProcessor) caches entry span, assuming only one per trace as before. API `set_transaction_name` uses cached span for direct `set_attribute` call.

There is another new class `SolarWindsEntrySpanProcessor` that's only for debugging at this point.

Tests failing. May need a cache cleanup function.

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -7,7 +7,6 @@
 import logging
 from typing import Any
 
-from opentelemetry import baggage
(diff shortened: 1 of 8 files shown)


tammy-baylis-swi:
Replaced by https://github.com/solarwinds/apm-python/pull/558

PR #558, opened 2025-03-27
NH-72398 Overhaul of default and SDK-set transaction naming for Pure Python and OTLP by default

tammy-baylis-swi:
_This will merge into epic feature branch_ `pure-python`.

Before this PR: APM used c-lib inbound metrics outputs and any `set_transaction_name` SDK calls to calculate transaction names. These were stored in a custom caching system (a name manager, a custom data type, multiple span processors, leverage of context baggage) then retrieved to write to events via `addInfo` at APM-proto export.

With this PR:
1. Transaction name calculation is pure Python, including cardinality limits via a new pool.
2. Upstream instrumentor-set span attributes (for HTTP, FAAS, etc) and default span name are used to directly set a default transaction name as span attribute at entry span `on_start` -- because ended Otel Python spans are immutable.
3. `on_start` also caches that entry span in Otel Context for access by SDK if called.
4. Any valid `set_transaction_name` SDK calls overwrite transaction name span attribute of the entry span in Context. This is always after span start and before span end.
5. At span `on_end`, the cached entry span is cleared.
6. Transaction name span attribute is set at time of default OTLP export.
7. The old custom caching system has been removed, as well as the Propagator hack to work around it.

Additional changes:
* AWS Lambda builds are updated to exclude wrapper completely
* CodeQL will run at push/PR for `pure-python` branch for dev convenience

See ticket comment for detailed test results. Please feel free to ask any questions or leave any ideas!


solarwinds_apm/version.py
```diff
@@ -1 +1 @@
-__version__ = "3.5.0"
+__version__ = "4.0.0.0"
```
jerrytfleung, line 1: Is it a typo?

jerrytfleung (approved):
left a nit comment

solarwinds_apm/version.py
tammy-baylis-swi: Nope, for staging/test publishes we use a 4th digit.

PR #562, opened 2025-04-02
NH-72398 Nit: Don't need to resolve `http.route` as txn name

tammy-baylis-swi:
We don't need to resolve `http.route` when used as default txn name because of how it's [defined in the semconv](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-route).


jerrytfleung (approved):
LGTM

PR #563, opened 2025-04-02
NH-104794 Simplify ResponseTime metrics, request counters for pure Python

tammy-baylis-swi:
Simplifies ResponseTime metrics generation and request counters for APM Pure Python.

`SolarWindsOTLPMetricsSpanProcessor` houses `response_time` histogram now. It continues to `record` at `on_end`.

Removes now-unnecessary MeterManager and OboeAPI usage from metrics processing because pure Python Sampler does request count calculation and `add`.


jerrytfleung (approved):
Code looks neat!

PR #564, opened 2025-04-03
NH-104796 Fix init of pure Python Oboe counters

tammy-baylis-swi:
Fixes small bug in the init of pure Python Oboe request counters for sampler-triggered OTLP metrics. Changes to use current global MeterProvider instead of passing a new one, so all metrics export via one reader etc.


jerrytfleung (approved):
Thanks for correction.

PR #565, opened 2025-04-03
NH-104999 Remove unused `oboe_api` from configurator, sampler init

tammy-baylis-swi:
A bit of cleanup along the way. Removes unused `oboe_api` from configurator, sampler init. Pure Python sampler in any mode does not need it.


jerrytfleung (approved):
Nice cleanup. Thanks!

PR #567, opened 2025-04-03
NH-104796 Rename as ResponseTimeProcessor

tammy-baylis-swi:
Renames APM Python `SolarWindsOTLPMetricsSpanProcessor` to simpler `ResponseTimeProcessor`. It's not so all-encompassing now that we've created the Pure Python sampler, and it's only responsible for recording `response_time` metrics.


jerrytfleung (approved):
LGTM. Thanks!

PR #568, opened 2025-04-03
NH-104999 Remove Inbound metrics span processor, merge base into ResponseTimeProcessor

tammy-baylis-swi:
Removes inbound metrics span processor because all request counter and response time metrics are generated in Pure Python. Merges previously shared base class into ResponseTimeProcessor class.


jerrytfleung (approved):
LGTM. Thanks!

PR #570, opened 2025-04-03
NH-104999 Remove now-unused ApmConfig helpers, constants

tammy-baylis-swi:
Removes some now-unused ApmConfig helpers functions and constants.

`INTL_SWO_AO_COLLECTOR` and `INTL_SWO_AO_STG_COLLECTOR` are default AO collector endpoints we no longer use for pure-Python/OTLP-export instrumentation.

`_calculate_metric_format` and `_calculate_certificates` are no longer called. They were used for legacy Reporter init.

`_calculate_logs_enabled` is no longer called. It was a safeguard if `SW_APM_EXPORT_LOGS_ENABLED` but collector was an AO endpoint.


jerrytfleung (approved):
LGTM

PR #572, opened 2025-04-04
NH-102995 No-op SolarWindsSpanExporter

tammy-baylis-swi:
No-ops the SolarWindsSpanExporter. Now it does nothing with finished Otel spans and exports nothing. If it's configured to load, then it logs a warning: 

`2025-04-04 17:26:01,154 [ solarwinds_apm.exporter WARNING  p#1.281473861460000] SolarWindsSpanExporter is no-op as of version 4.0.0 and will be completely removed in a future release. Please update configuration to use OTel Python HTTP or gRPC OTLP exporter.`

Removes no-longer-used constants. Simplifies SolarWindsSpanExporter init to take fewer args, so simplifies the Configurator as well. Removes `SolarWindsFrameworkKvManager` which existed to optimize setting `Python.*.Version` on oboe events (no longer needed).

I will be adjusting exporter defaults/fall-backs in a different PR.


jerrytfleung (approved):
LGTM

PR #573, opened 2025-04-05
NH-106956 OTLP exporter compression gzip by default

tammy-baylis-swi:
APM Python sets [OTLP exporter compression](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables/__init__.py#L315-L330) to `"gzip"` by default. User can still configure compression to `"deflate"` if they wish.


cheempz (approved):

PR #574, opened 2025-04-05
NH-106957 All meters under APM Python use Delta as preferred default

tammy-baylis-swi:
All meters initialized under APM Python will use Delta as preferred default for aggregation temporality, not just histograms.


cheempz (approved):

PR #575, opened 2025-04-07
NH-107152 Fix image builds

tammy-baylis-swi:


cheempz (approved):

PR #578, opened 2025-04-07
NH-101363 Configurator uses more upstream logic

tammy-baylis-swi:
tl;dr Updates APM custom Configurator to use more of upstream OTel's logic.

`SolarWindsConfigurator` was already extending `_OTelSDKConfigurator` but was completely overriding the `_configure` function. It actually still has to with this update in order to customize certain settings when [_init_tracing](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L194) and [_init_metrics](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L214) are not instance/class functions. But now it also calls upstream's `_import_exporters` etc to stop duplicating that logic. Configured component validity is now checked by upstream, e.g. [this check of SpanExporter ](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L302-L305). Detecting configuration by `OTEL_*` environment variable is now done by upstream, i.e. [this check of OTEL_*_EXPORTER](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L183).
		
`SolarWindsConfigurator` needs `_custom_init_metrics` and `_custom_init_tracing` to support:

1. custom TracerProvider with sampler shutdown
2. Batch vs Simple span processor init, for lambda
3. Metrics export interval default vs instant, for lambda
4. custom default aggregation temporality

The `agent_enabled` check is done once at the start of `_configure` now. If false, all tracer/meter/logger providers are no-op'd and none of the sampler, span processors, propagators are initialized. The ApmConfig class still does calculation of `agent_enabled` based on SW APM config.

Re. unit test changes: Completely removes some unit test classes because the responsibilities those cases tested are now the responsibility of upstream or just removed/refactored. Examples: (1) `_init_logging` instead of now-removed `_configure_log_exporter`. (2) top-level `_configure` inits the sampler instead of now-removed `_configure_sampler`. Repurposes and renames `test_configurator_(traces|metrics)_init.py` to test new custom methods for APM/Lambda-specific settings.

See manual test results on ticket comment.

Please let me know if any questions or comments!


jerrytfleung (approved):
LGTM.

PR #580, opened 2025-04-10
NH-98561 Update distro setdefaults to pure OTLP HTTP

tammy-baylis-swi:
### Summary of changes

APM Python defaults all of traces, metrics, logs export to direct SWO by OTLP HTTP, with service key token automatically set in export headers.

If user manually configures `OTEL_TRACES_EXPORTER` to the now no-op `solarwinds_exporter`, APM logs a warning then sets either OTLP HTTP or gRPC exporter for span exporter, depending on set or default OTLP protocol.

### Details

<details>
<summary>Click to expand!!!!!!</summary>

We stop setting APM-proto as the default traces exporter. The default is now the upstream OTLP HTTP exporter. OTLP gRPC can be configured by user. I'm choosing HTTP as the default because it's required in k8s environment, and doesn't harm other use cases.
		
We no longer need to support signal-specific defaults because there is no longer a split between traces (APM-proto default) and metrics/logs (OTLP default but only if protocol set). We set general defaults and allow upstream components to map general to signal-specific where possible.

Any signal-specific protocol and its precedence is respected by [this upstream configurator helper](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L146-L149).

Any signal-specific export endpoint and its precedence is respected by upstream exporters, [for example](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/exporter/opentelemetry-exporter-otlp-proto-grpc/src/opentelemetry/exporter/otlp/proto/grpc/trace_exporter/__init__.py#L126-L127).

We don't subclass [OpenTelemetryDistro](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/6d5a5149d02fdec50d8053767b55bc91d80a9228/opentelemetry-distro/src/opentelemetry/distro/__init__.py#L31) since (a) it doesn't do much at the moment, (b) it would introduce `opentelemetry-distro` as a dependency, and (c) it would require working around the entry points where OTelDistro and OTelConfigurator are loaded as defaults and not the APM versions.

Any valid protocol (default http/protobuf, or grpc) will be mapped by configurator to matching exporters, if exporters not configured.

If protocol and exporters conflict, then [this upstream helper](https://github.com/open-telemetry/opentelemetry-python/blob/3644a1e17915aa7e8820dbacc1402338f9bfbb1b/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L167-L175) logs that warning and uses exporter anyway.

Fewer AWS Lambda wrapper default overrides (but still defaults) are required.
</details>

### Testing

Several unit tests added/updated. See also ticket comment for manual testing results.


lambda/solarwinds-apm/wrapper
```diff
@@ -27,16 +27,6 @@ export LAMBDA_LAYER_PKGS_DIR="/opt/python";
 export PYTHONPATH="$LAMBDA_LAYER_PKGS_DIR:$PYTHONPATH";
 export PYTHONPATH="$LAMBDA_RUNTIME_DIR:$PYTHONPATH";
 
-# Default opt into OTLP metrics export in lambda
-if [ -z "${SW_APM_EXPORT_METRICS_ENABLED}" ]; then
-    export SW_APM_EXPORT_METRICS_ENABLED=true
-fi
```
tammy-baylis-swi, line 33: This should have been removed in a different PR where `true` is now the default.

jerrytfleung (approved):
LGTM

PR #581, opened 2025-04-10
NH-101363 Remove OTEL_TRACES_EXPORTER check when calculating agent_enabled

tammy-baylis-swi:
Removes the `OTEL_TRACES_EXPORTER` validity check that was part of calculating `agent_enabled`. Now that we let upstream handle configuration support (see https://github.com/solarwinds/apm-python/pull/578), APM side shouldn't duplicate or do something similar anymore.


jerrytfleung (approved):
LGTM. More upstream please!

PR #582, opened 2025-04-11
NH-104961 - Python: Pure Python WaitUntilReady SDK

jerrytfleung:
- Implemented solarwinds_ready api function
- Added tests

https://swicloud.atlassian.net/browse/NH-104961

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -10,18 +10,15 @@ from typing import Any
 from opentelemetry import context
 from opentelemetry.trace import NoOpTracerProvider, get_tracer_provider
 
-# pylint: disable=import-error,no-name-in-module
-# from solarwinds_apm.extension.oboe import Context
 from solarwinds_apm.apm_constants import (
     INTL_SWO_OTEL_CONTEXT_ENTRY_SPAN,
     INTL_SWO_TRANSACTION_NAME_ATTR,
 )
 from solarwinds_apm.oboe import get_transaction_name_pool
+from solarwinds_apm.oboe.http_sampler import HttpSampler
 from solarwinds_apm.oboe.transaction_name_pool import TRANSACTION_NAME_DEFAULT
+from solarwinds_apm.tracer_provider import SolarwindsTracerProvider
 from solarwinds_apm.w3c_transformer import W3CTransformer
-
-# from solarwinds_apm.apm_oboe_codes import OboeReadyCode
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


solarwinds_apm/tracer_provider.py
```diff
@@ -11,6 +11,12 @@
 
 
 class SolarwindsTracerProvider(TracerProvider):
+    def get_sampler(self):
+        """
+        Returns the current sampler.
+        """
+        return self.sampler
```
tammy-baylis-swi, line 18: Since it's not `self._sampler` in parent class, then it would be fine to call `sampler = tracer_provider.sampler` instead of `sampler = tracer_provider.get_sampler()`. But an explicit getter is nice so this is fine.

solarwinds_apm/api/__init__.py
```diff
@@ -97,45 +95,33 @@ def set_transaction_name(custom_name: str) -> bool:
     return True
 
 
-def solarwinds_ready(
-    wait_milliseconds: int = 3000,
-    integer_response: bool = False,
-) -> Any:
+def solarwinds_ready(wait_milliseconds: int = 3000) -> Any:
```
tammy-baylis-swi, line 98:
```suggestion
def solarwinds_ready(wait_milliseconds: int = 3000) -> bool:
```

solarwinds_apm/api/__init__.py
```diff
@@ -97,45 +95,33 @@ def set_transaction_name(custom_name: str) -> bool:
...
+    if isinstance(tracer_provider, SolarwindsTracerProvider):
+        sampler = tracer_provider.get_sampler()
+        if isinstance(sampler, (HttpSampler, JsonSampler)):
+            return sampler.wait_until_ready(int(wait_milliseconds / 1000))
+        logger.debug(
+            "SolarWinds not ready because sampler is not a Solarwinds-specific sampler"
+        )
+    logger.debug(
+        "SolarWinds not ready because tracer_provider is not a SolarwindsTracerProvider"
+    )
```
tammy-baylis-swi, line 126: The `isinstance` check above makes sense. Maybe this message could be updated to help with deeper debugging. If it's upstream regular TracerProvider then our config is wrong. If ProxyTracerProvider then there could be something wrong with auto-instrumentation.

```suggestion
    logger.debug(
        "SolarWinds not ready. Got APM TracerProvider: %s",
        type(tracer_provider),
    )
```

tammy-baylis-swi (approved):
Thanks Lgtm! Just some nits. I'll create a ticket to update docs.

solarwinds_apm/tracer_provider.py
jerrytfleung: Sure. Done

PR #584, opened 2025-04-14
NH-107698 Python: Add back solarwinds_ready integer_response arg as no-op

jerrytfleung:
Add back integer_response arg for compatibility.
https://swicloud.atlassian.net/browse/NH-107698


tammy-baylis-swi (approved):
Lgtm, thanks!

PR #589, opened 2025-04-14
NH-107706 Python: Pure Python default collector

jerrytfleung:
- Fixed unset SW_APM_COLLECTOR yield to a `None` value from `apm_config['collector']` issue
- Added unit test case

https://swicloud.atlassian.net/browse/NH-107706


tammy-baylis-swi (approved):
Lgtm, good find! I've checked this out and was able to auto-instrument a service without specifying collector now 👍

jerrytfleung:
> Lgtm, good find! I've checked this out and was able to auto-instrument a service without specifying collector now 👍

Yes, it is also tested in the unit test

PR #590, opened 2025-04-15
NH-102467 Python: Add deducing of export mode from SW_APM_COLLECTOR

jerrytfleung:
Added deducing mode from SW_APM_COLLECTOR

https://swicloud.atlassian.net/browse/NH-102467

### DIFF
--- a/tests/unit/test_distro.py
+++ b/tests/unit/test_distro.py
@@ -346,6 +346,77 @@ class TestDistro:
         assert os.environ[OTEL_LOGS_EXPORTER] == "otlp"
         assert os.environ.get("OTEL_SEMCONV_STABILITY_OPT_IN") == "http"
 
+    def test_configure_env_service_key_only(self, mocker):
+        mocker.patch.dict(
+            os.environ,
+            {
+                "SW_APM_SERVICE_KEY": "foo-token:bar",
+            },
+            clear=True
+        )
+        distro.SolarWindsDistro()._configure()
+        assert os.environ[OTEL_EXPORTER_OTLP_ENDPOINT] == "https://otel.collector.na-01.cloud.solarwinds.com:443"
+
+    def test_configure_env_service_key_only(self, mocker):
+        mocker.patch.dict(
+            os.environ,
+            {
+                "SW_APM_SERVICE_KEY": "foo-token:bar",
+            },
+            clear=True
+        )
+        distro.SolarWindsDistro()._configure()
+        assert os.environ[OTEL_EXPORTER_OTLP_ENDPOINT] == "https://otel.collector.na-01.cloud.solarwinds.com:443"
+
+    def test_configure_env_service_key_and_collector_only(self, mocker):
+        mocker.patch.dict(
+            os.environ,
+            {
+                "SW_APM_SERVICE_KEY": "foo-token:bar",
+                "SW_APM_COLLECTOR": "apm.collector.na-02.cloud.solarwinds.com",
+            },
+            clear=True
+        )
+        distro.SolarWindsDistro()._configure()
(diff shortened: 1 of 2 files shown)


tests/unit/test_distro.py
```diff
@@ -346,6 +346,78 @@ def test_configure_env_endpoint(self, mocker):
...
+        )
+        distro.SolarWindsDistro()._configure()
+        assert os.environ[OTEL_EXPORTER_OTLP_ENDPOINT] == "https://otel.collector.na-02.st-ssp.solarwinds.com:443"
+
+    def test_configure_env_service_key_and_collector_only_invalid(self, mocker):
+        mocker.patch.dict(
+            os.environ,
+            {
+                "SW_APM_SERVICE_KEY": "foo-token:bar",
+                "SW_APM_COLLECTOR": "www.google.com",
```
tammy-baylis-swi, line 401: This one I am not sure about, and apologies because it's not in ticket OP either.

@cheempz , does this make sense for this iteration?: If `SW_APM_COLLECTOR` is anything but `apm.collector`, then APM uses the `otel.collector.na-01` default. OR, APM uses what's configured and lets settings collection and export fail for visible error.

tests/unit/test_distro.py
```diff
@@ -346,6 +346,78 @@ def test_configure_env_endpoint(self, mocker):
...
+            },
+            clear=True
+        )
+        distro.SolarWindsDistro()._configure()
+        assert os.environ[OTEL_EXPORTER_OTLP_ENDPOINT] == "https://otel.collector.na-01.cloud.solarwinds.com:443"
+
```
tammy-baylis-swi, line 407: Weird ask because it's repetitive but it'll help with spec clarity:

Could we have an additional unit test where `"SW_APM_COLLECTOR": "otel.collector.na-02.st-ssp.solarwinds.com"` then the assert is the endpoint is the default `"https://otel.collector.na-01.cloud.solarwinds.com:443"`. This is to make clear that `COLLECTOR` doesn't get resolved if `otel.collector` which I _think_ is how we address that edge case.

tammy-baylis-swi:
Thank you! Overall looks good but I'd like a 2nd set of eyes on this before merging since it relates to this week's conversations 🙂

tests/unit/test_distro.py
jerrytfleung: Sure. Done

tests/unit/test_distro.py
jerrytfleung: My thinking was to greedily extract `xx-yy.<cloud/st-ssp/dev-ssp>` from the environment variable and substitute to `https://otel.collector.<substitute>.solarwinds.com:443`

Let me know if it doesn't making sense. Thanks!

tammy-baylis-swi (approved):

tests/unit/test_distro.py
cheempz: yes it makes sense to use the fallback value if SW APM config for collector is anything other than `apm.collector.*`

tests/unit/test_distro.py
```diff
@@ -346,6 +346,90 @@ def test_configure_env_endpoint(self, mocker):
...
+    def test_configure_env_service_key_and_collector_and_otel_exporter(self, mocker):
+        mocker.patch.dict(
+            os.environ,
+            {
+                "SW_APM_SERVICE_KEY": "foo-token:bar",
+                "SW_APM_COLLECTOR": "apm.collector.na-02.cloud.solarwinds.com",
+                "OTEL_EXPORTER_OTLP_ENDPOINT": "https://apm.collector.na-03.cloud.solarwinds.com:443",
```
cheempz, line 426: this is a good test that we respect explict OTEL exporter config as highest precedence. my nit is the value could be something that does not imply the apm.collector actually supports OTLP.

jerrytfleung: Sure. Done

PR #592, opened 2025-04-15
NH-107791 Configurator span exporter loading with endpoints instead of grpcio importing

tammy-baylis-swi:
Changes the Configurator to check if `solarwinds_exporter` has been configured with new helper `_swap_legacy_span_exporter` that tweaks the return of `_get_exporter_names("traces")`, instead of direct gRPC exporter export at `_custom_init_tracing`. This is to prevent `ImportError` of `grpcio` in some environments. I've chosen this way because overriding more upstream configuration methods would be too duplicative.

Like https://github.com/solarwinds/apm-python/pull/580: if user manually configures `OTEL_TRACES_EXPORTER` to the now no-op `solarwinds_exporter`, APM logs a warning then sets either OTLP HTTP or gRPC exporter for span exporter, depending on set or default OTLP protocol.

Please let me know your thoughts!


jerrytfleung (approved):
The test cases covered 100% of `_swap_legacy_span_exporter` function!

LGTM. Thanks!

PR #577, opened 2025-04-07
NH-103814 - Python: use resource attributes for SWO entity inference

jerrytfleung:
- Added dependency to use resource detectors for aws, azure
- Added k8s, uams resource detector implementation
- Added unit tests
- Added default environment to enable resource detectors in apm-python

Manual testing results and setup
https://swicloud.atlassian.net/wiki/spaces/NIT/pages/5479661616/2025-04-08+Resource+detector+uams+k8s

https://swicloud.atlassian.net/browse/NH-103814

PR #593, opened 2025-04-15
NH-107714 lambda resource detect

tammy-baylis-swi:
Updates Lambda dependencies and specifies resource detectors for Lambda environment.

Expands on https://github.com/solarwinds/apm-python/pull/577. See ticket comment for some testing.

### DIFF
--- a/lambda/requirements.txt
+++ b/lambda/requirements.txt
@@ -1,6 +1,7 @@
 opentelemetry-exporter-otlp==1.31.1
 opentelemetry-exporter-otlp-proto-http==1.31.1
 opentelemetry-instrumentation-aws-lambda==0.52b1
+opentelemetry-sdk-extension-aws==2.1.0
 packaging
 jaraco.text
 platformdirs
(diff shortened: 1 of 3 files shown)


lambda/requirements.txt
```diff
@@ -1,6 +1,7 @@
 opentelemetry-exporter-otlp==1.31.1
 opentelemetry-exporter-otlp-proto-http==1.31.1
 opentelemetry-instrumentation-aws-lambda==0.52b1
+opentelemetry-sdk-extension-aws==2.1.0
```
tammy-baylis-swi, line 4: I don't think we need Azure resource detector, so not included.

solarwinds_apm/apm_constants.py
```diff
@@ -40,6 +40,12 @@
     "k8s",
     "uams",
 ]
+INTL_SWO_DEFAULT_RESOURCE_DETECTORS_LAMBDA = [
```
tammy-baylis-swi, line 43: Please let me know if more/less should be part of this, like `INTL_SWO_DEFAULT_RESOURCE_DETECTORS`

jerrytfleung: I think the list is good.

jerrytfleung (approved):
LGTM. Thanks!

PR #594, opened 2025-04-15
NH-107709 Simplify lambda builds with Pure Python APM

tammy-baylis-swi:
Simplifies APM lambda layer builds now that APM is pure python, and additional changes from https://github.com/solarwinds/apm-python/pull/591, https://github.com/solarwinds/apm-python/pull/592.

Layer publishes are now no longer split by architecture! The first major release will be `arn:aws:lambda:<REGION>:851060098468:layer:solarwinds-apm-python-4_0_0:1`

The Makefile has been updated to only pre-install lambda deps with Python 3.9. They are compatible with all of Python 3.8-3.12 in both `x86_64, arm64` because we no longer need to accommodate for `.so` of certain libraries.

See ticket comment for detailed testing.


Makefile
```diff
@@ -185,31 +185,21 @@ package: sdist check-sdist-local manylinux-wheels check-wheel-local
...
+	@if [ -f ./dist/solarwinds_apm_lambda.zip ]; then \
+		echo -e "Deleting old solarwinds_apm_lambda.zip"; \
+		rm ./dist/solarwinds_apm_lambda.zip; \
 	 fi
 	rm -rf ${target_dir}
 	@echo -e "Creating target directory ${target_dir} for AWS Lambda layer artifacts."
 	mkdir -p ${target_dir}/python
 	@echo -e "Install setuptools"
-	@set -e; for PYBIN in cp38-cp38 cp39-cp39 cp310-cp310 cp311-cp311 cp312-cp312; do /opt/python/$${PYBIN}/bin/pip install setuptools; done
+	/opt/python/cp312-cp312/bin/pip install setuptools
```
jerrytfleung, line 196: Is it just a name and doesn't bind to python 3.12?

jerrytfleung (approved):
Just left a question.

The PR looks good to me.

Makefile
tammy-baylis-swi: This does bind to Python 3.12 to do the pip install. Else it would use the container's default Python 3.8, which is probably fine but OTel Python will drop Python 3.8 support soon.

tammy-baylis-swi: Changed to Python 3.9 as discussed.

tammy-baylis-swi:
Thanks @jerrytfleung ; please have another look.

jerrytfleung (approved):
LGTM. Thanks! @tammy-baylis-swi

PR #595, opened 2025-04-16
NH-107710 Update install tests for pure python APM pypi builds

tammy-baylis-swi:
Updates the APM Python install tests / smoke tests for pure Python builds. Removes c-lib extension file checks. Changes the startup logs asserts because SSL Reporter is no longer used and we do http getSettings instead.

Successful run after installing [TestPyPI 4.0.0.2](https://test.pypi.org/project/solarwinds-apm/4.0.0.2/#files), an early test pure Python build: https://github.com/solarwinds/apm-python/actions/runs/14504062668


.github/workflows/verify_install.yaml
```diff
@@ -23,10 +23,8 @@ on:
 
 env:
   SOLARWINDS_APM_VERSION: ${{ github.event.inputs.solarwinds-version }}
-  SW_APM_COLLECTOR_AO_PROD: ${{ secrets.SW_APM_COLLECTOR_AO_PROD }}
```
tammy-baylis-swi, line 26: These secrets are not used anywhere else, so I have deleted them from the repo.

jerrytfleung (approved):
LGTM. Thanks!

PR #596, opened 2025-04-16
NH-103814 Don't load LambdaDetector when outside lambda

tammy-baylis-swi:
Don't load LambdaDetector when outside lambda, otherwise upstream gives us a warning: `AwsLambdaResourceDetector failed: 'AWS_REGION'`.


jerrytfleung (approved):
LGTM. Thanks for updating the list.

PR #597, opened 2025-04-16
NH-107710 PyPI publish without separate aarch64 build

tammy-baylis-swi:
Removes the splits in PyPI / TestPyPI builds of APM Python so that it's a single sdist + wheel publish for Pure Python.

[Used to publish](https://github.com/solarwinds/apm-python/actions/runs/14504959039) testrelease 4.0.0.5: https://test.pypi.org/project/solarwinds-apm/4.0.0.5/#files. The future PyPI publish should be the same.


jerrytfleung (approved):
Great! LGTM. 
The workflow is now simplified.

PR #600, opened 2025-04-17
NH-107912 Stop logging that Lambda mode experimental

tammy-baylis-swi:
Stop logging that Lambda mode experimental!


jerrytfleung (approved):
LGTM

PR #602, opened 2025-04-21
NH-108028 Support config file collector endpoint resolution

tammy-baylis-swi:
Adds new ApmConfig class function to calculate `collector` as `environment Variable > config file > default`. Distro now uses this for `collector` resolution of export endpoint. This is needed because Configurator (and ApmConfig) init happen after Distro lifecycle has ended; Distro and Configurator can't share one ApmConfig due to lifecycles when [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/748c92592d2f476199667629defce4a3bca9ecc9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L37-L39) is used.

### DIFF
--- a/solarwinds_apm/apm_config.py
+++ b/solarwinds_apm/apm_config.py
@@ -124,6 +124,25 @@ class SolarWindsApmConfig:
         return False
 
     @classmethod
+    def calculate_collector(
+        cls,
+        cnf_dict: dict = None,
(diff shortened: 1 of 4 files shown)


jerrytfleung (approved):
LGTM.

PR #604, opened 2025-04-21
NH-104807 - Python Reset warning message of http sampler

jerrytfleung:
Reset last warning message if http_sampler recovered itself.

Found this issue during porting code from python to PHP

https://swicloud.atlassian.net/browse/NH-104807


tammy-baylis-swi (approved):
Thanks for catching and porting back!

PR #605, opened 2025-04-22
NH-105347 Update test scenarios for Pure Python

tammy-baylis-swi:
Comments all integration tests back in, and updates them to test Pure Python APM. The setup is updated so that the test scenarios use `JsonSampler` and a mock to guarantee sampling decision. Assertions are almost all the same, with these updates:

1. Sampler more correctly sets `BucketCapacity` and other SW KVs -- legacy was sometimes setting `-1` unnecessarily.
2. Sampler does not cache any `sw` KV in TraceState at time of should sample, like legacy did. Setting `sw` only needs to be done by custom Propagator.inject and this is already the case.

Includes small bugfix of ApmConfig for case where default service name [_starts with_ "unknown_service".](https://github.com/open-telemetry/opentelemetry-python/blob/f5fb6b1353929cf8039b1d38f97450866357d901/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L174-L183)

Includes small bugfix of x-trace-options response header piggyback sanitization.

Includes refactor of `JsonSampler._loop` for easier mocking.


tests/integration/test_scenario_1.py
```diff
@@ -1,137 +1,156 @@
...
+           (done by OTel TraceContextTextMapPropagator).
+        4. Sampling-related attributes are set for the root/service entry span.
+        5. The span_id of the outgoing request span matches the span_id portion in the
+           tracestate header.
+        """
+        # Use in-process test app client and mock to propagate context
+        # and create in-memory trace
+        resp = None
+        # Mock JSON read to guarantee sample decision
+        timestamp = time.time()
```
jerrytfleung, line 38: need integer casting. e.g. timestamp = int(time.time())

tammy-baylis-swi: Addressed in [645909a](https://github.com/solarwinds/apm-python/pull/605/commits/645909a8b9b1db9129409c7d01e825a99908723f)

jerrytfleung (approved):
LGTM

PR #606, opened 2025-04-22
NH-108175 Python: Fix sw.w3c.tracestate attribute

jerrytfleung:
sw.w3c.tracestate is not found in the span attributes.
It is because a wrong trace_state was used to check for the inbound trace_state. As a result, the wrong trace state yield a None and skip the attributes setup

Fixed by checking the trace_state from parent_span_context

https://swicloud.atlassian.net/browse/NH-108175


tammy-baylis-swi (approved):
Lgtm, thanks for updated tests as well.

PR #608, opened 2025-04-24
NH-107450 Add units, description to OTel counters for metrics

tammy-baylis-swi:
Adds units and descriptions to OTel counters for our metrics.

Already being added for the histogram: https://github.com/solarwinds/apm-python/blob/db06001fb4ea3570b0bfc9409d08d431fa3742dc/solarwinds_apm/trace/response_time_processor.py#L49-L53


jerrytfleung (approved):
LGTM

PR #609, opened 2025-04-24
NH-104999 Remove unnecessary param at Pure Python sampler init

tammy-baylis-swi:
Removes now-unnecessary param, `reporter`, at Pure Python sampler init. Re-adds a unit test.


jerrytfleung (approved):
LGTM

PR #610, opened 2025-04-24
NH-104999 Remove legacy sampler tests

tammy-baylis-swi:
Depends on https://github.com/solarwinds/apm-python/pull/609

Removes unit tests and fixtures that are no longer relevant after decommissioning of legacy liboboe sampler.

Additional rationale: All the `x-trace-options` response header logic, including the "piggyback", is in OboeSampler and tested in `test_oboe_sampler.py` and the integration tests. Same for calculating span attributes with our different `*_algo` functions, calculating transaction name with the pool, and tracing mode. All parent vs child sampling logic is managed and tested by upstream, with additional testing in our integration tests. There is no more c-lib extension-proxied logic like `calculate_liboboe_decision`, `otel_decision_from_liboboe`.


jerrytfleung (approved):
LGTM. Says bye to liboboe!

PR #614, opened 2025-04-25
NH-104999 Fully remove extension from linting, builds

tammy-baylis-swi:
Fully remove extension mentions from linting, builds.

### DIFF
--- a/solarwinds_apm/api/__init__.py
+++ b/solarwinds_apm/api/__init__.py
@@ -33,7 +33,7 @@ def set_transaction_name(custom_name: str) -> bool:
     config file.
 
     Any uppercase to lowercase conversions or special character replacements
-    are done by the platform. Any truncations are done by the core extension.
+    are done by the platform.
 
     :custom_name:str, custom transaction name to apply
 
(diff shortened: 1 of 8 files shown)


solarwinds_apm/api/__init__.py
```diff
@@ -33,7 +33,7 @@ def set_transaction_name(custom_name: str) -> bool:
     config file.
 
     Any uppercase to lowercase conversions or special character replacements
-    are done by the platform. Any truncations are done by the core extension.
+    are done by the platform.
```
tammy-baylis-swi, line 36: Does any truncation happen without core extension?

(Just found this while grepping)

jerrytfleung: Yes, we allow max 256 characters of a transaction name. It is in the transaction_name_pool.py

jerrytfleung (approved):
LGTM

solarwinds_apm/api/__init__.py
tammy-baylis-swi: Thanks! I updated it in [a6ed301](https://github.com/solarwinds/apm-python/pull/614/commits/a6ed3013607bca3d09171a1be0aec33c52810409) to

```
    are done by the platform. Name length is limited to 256 characters;
    anything longer is truncated by APM library.
```

PR #616, opened 2025-04-25
NH-23784 NH-104999 Build sdist without setup.py; use build for sdist

tammy-baylis-swi:
Updates sdist builds to no longer need setup.py, which is easier since APM Python is pure Python. We do this because setup.py is legacy and pyproject.toml should be used: https://pip.pypa.io/en/stable/reference/build-system/#build-system-interface. Adds `build` dependency to publish/dev container for building sdist.

TestPyPI was published successfully to https://test.pypi.org/project/solarwinds-apm/4.0.0.10/#files and smoke tests pass: https://github.com/solarwinds/apm-python/actions/runs/14674509169. See ticket comment for additional testing.


jerrytfleung (approved):
LGTM

PR #633, opened 2025-05-10
NH-108983 Adjust "Verify Installation" test dependencies following Pure Python

tammy-baylis-swi:
Adjusts "Verify Installation" test dependencies following Pure Python. Generally the deps of each Linux distro can be reduced to "regular" Python and utils to curl/find/grep etc, without needing "dev" Python and g++.

We do add an Alpine dependency, `linux-headers`, needed since adding psutils dep in https://github.com/solarwinds/apm-python/pull/629. This will be needed after next release of APM Python.

Test run using this branch and APM Python 4.0.0: https://github.com/solarwinds/apm-python/actions/runs/14940259260


jerrytfleung (approved):
LGTM

PR #634, opened 2025-05-12
NH-110601 - Python: Fix function signature

jerrytfleung:
Fixed function signature

https://swicloud.atlassian.net/browse/NH-110601


tammy-baylis-swi (approved):
Thanks for catching!

PR #635, opened 2025-05-13
NH-108983 Update "Verify Installation" CI/CD for more automation

tammy-baylis-swi:
Updates "Verify Installation" CI/CD to be more automated. A `"local"` registry/source is now supported and is the default, which uses contents of the current branch to build sdist and wheel to do the test (not just PyPI/TestPyPI anymore). This allows us to automatically run "Verify Installation" with the other tox tests now on every PR and merge to `main` branch.

This branch was used to successfully run install tests from "local" (branch contents) by manual dispatch: https://github.com/solarwinds/apm-python/actions/runs/14985714435

And they also ran on this PR 😺

### DIFF
--- a/tests/docker/install/_helper_run_install_tests.sh
+++ b/tests/docker/install/_helper_run_install_tests.sh
@@ -30,8 +30,8 @@ echo "Installing test dependencies for Python $python_version on $pretty_name"
     if grep Alpine /etc/os-release; then
         # test deps
         apk add bash
-        # agent deps - APM dep on psutil, so Alpine need linux-headers
-        apk add python3 curl linux-headers
+        # agent deps - APM deps on psutil, so Alpine needs more deps
+        apk add python3 curl linux-headers gcc musl-dev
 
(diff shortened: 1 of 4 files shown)


tests/docker/install/_helper_run_install_tests.sh
```diff
@@ -30,8 +30,8 @@ echo "Installing test dependencies for Python $python_version on $pretty_name"
     if grep Alpine /etc/os-release; then
         # test deps
         apk add bash
-        # agent deps - APM dep on psutil, so Alpine need linux-headers
-        apk add python3 curl linux-headers
+        # agent deps - APM deps on psutil, so Alpine needs more deps
+        apk add python3 curl linux-headers gcc musl-dev
```
tammy-baylis-swi, line 34: We have to add more deps back 🙃 but only for Alpine, in order to install dep `psutil`. See also https://github.com/solarwinds/apm-python/pull/633

PR #636, opened 2025-05-13
NH-108983 Add verify_installation_windows CI/CD

tammy-baylis-swi:
Adds `verify_installation_windows` CI/CD to smoke test the installation of APM Python on `windows-latest` for the 5 supported Python versions. This will be run on every PR and push to main with current branch contents, and can be manually triggered to install from local (branch contents), TestPyPI, or PyPI -- like the update in https://github.com/solarwinds/apm-python/pull/635. This re-writes the existing `.sh` test scripts as `.ps1` for Windows compatibility.

Completed test runs:
* "local" (with correct job names!): https://github.com/solarwinds/apm-python/actions/runs/15008808953
* "testpypi": https://github.com/solarwinds/apm-python/actions/runs/15008683418/job/42173195998
* "pypi": https://github.com/solarwinds/apm-python/actions/runs/15008722309

PR #637, opened 2025-05-14
NH-108983 Add verify_installation_macos CI/CD

tammy-baylis-swi:
Adds `verify_installation_macos` CI/CD to smoke test the installation of APM Python on `macos-latest` for the 5 supported Python versions. This will be run on every PR and push to main with current branch contents, and can be manually triggered to install from local (branch contents), TestPyPI, or PyPI -- like the updates in https://github.com/solarwinds/apm-python/pull/635 and https://github.com/solarwinds/apm-python/pull/636. MacOS is Unix-based so we use the existing Linux test scripts with minor adjustments.

Completed test run: https://github.com/solarwinds/apm-python/actions/runs/15009592518


jerrytfleung (approved):
LGTM. 

Indeed I was thinking is it possible to combine all platform into a single verify_install.yaml where the matrix covers all platforms.

PR #638, opened 2025-05-15
NH-110883 psutil becomes optional dep for APM Python

tammy-baylis-swi:
`psutil` is now an optional dependency of APM Python. This isn't breaking because the initial addition of `psutil` has not yet been released.

From user point of view, system metrics generation is now an opt-in at time of installation:
* "Regular" install: `pip install solarwinds-apm`
* Opt into metrics generation: `pip install solarwinds-apm[system-metrics]`

I've updated the install tests to always opt in. Successful test runs:
- Linux https://github.com/solarwinds/apm-python/actions/runs/15055786278
- Mac https://github.com/solarwinds/apm-python/actions/runs/15055789490
- Windows https://github.com/solarwinds/apm-python/actions/runs/15055791947

I'm working on a separate PR for documentation update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ pip install solarwinds-apm
 opentelemetry-bootstrap --action=install
 ```
 
+To opt into system metrics generation, specify `pip install solarwinds-apm[system-metrics]`.
+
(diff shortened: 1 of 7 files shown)


jerrytfleung (approved):
LGTM.

PR #639, opened 2025-05-15
Updated description

jerrytfleung:
Nits

- Updated counter description as there are other factors to not sample the request ( e.g. the dice returned unsampled)


tammy-baylis-swi (approved):
Thanks!

cheempz:
Oh hey just saw this one... can the description be standardized to same exact value as listed in https://swicloud.atlassian.net/browse/NH-106169, to avoid flip-flopping in the UI.

PR #641, opened 2025-05-16
NH-110883 Remove psutil dependency

tammy-baylis-swi:
Remove `psutil` dependency -- not breaking because APM Python has not been released with the psutil dep.

Instead, we instruct users to install psutil separately to opt into system metrics export.


cheempz (approved):

PR #642, opened 2025-05-16
NH-111009 Remove system-metrics instrumentor pre-installation from k8s

tammy-baylis-swi:
Remove the system-metrics instrumentor pre-installation from k8s image publishes. This is a precaution against some performance and money penalties.


cheempz (approved):
👍

PR #666, opened 2025-07-15
NH-110318 Configurator sets Resource with service.instance.id

tammy-baylis-swi:
Configurator sets Resource with `service.instance.id`.

### DIFF
--- a/solarwinds_apm/configurator.py
+++ b/solarwinds_apm/configurator.py
@@ -9,6 +9,7 @@
 import logging
 import math
 import os
+import uuid
 from typing import Optional, Union
 
 from opentelemetry import trace
@@ -170,6 +171,7 @@ class SolarWindsConfigurator(_OTelSDKConfigurator):
                 "sw.apm.version": __version__,
                 "sw.data.module": "apm",
                 "service.name": self.apm_config.service_name,
+                "service.instance.id": str(uuid.uuid4()),
             }
         )
         self._custom_init_tracing(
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM, thanks @tammy-baylis-swi! I'm assuming if a detector like Azure AppService sets the service instance id, it would overwrite this UUID based default.

tammy-baylis-swi:
> LGTM, thanks @tammy-baylis-swi! I'm assuming if a detector like Azure AppService sets the service instance id, it would overwrite this UUID based default.

@cheempz the opposite will happen. 🙃 The SDK's Resource [merge](https://github.com/open-telemetry/opentelemetry-python/blob/14401e1dd82852fdea76ecf4d720fbd74834b705/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L258-L276) is [called](https://github.com/open-telemetry/opentelemetry-python/blob/14401e1dd82852fdea76ecf4d720fbd74834b705/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L229-L231) using aggregated resources from the detectors as the "old" resource and the attributes we pass to `Resource.create` as the "new" ones. So the APM-side UUID would always used. I'll make some changes.

tammy-baylis-swi:
Updated! @cheempz please have a new look.

cheempz (approved):
Nice! thanks for the revisit and tests @tammy-baylis-swi!

PR #667, opened 2025-07-16
NH-115253 Add more Python 3.13 testing, update Lambda tags

tammy-baylis-swi:
Updates APM Python unit tests and installation/smoke tests to include Python 3.13 -- required some helper shell script changes. All pass on this PR's ci/cd.

Removes explicit Python version `Classifiers` from pyproject.toml to match upstream. We still have [requires-python](https://github.com/solarwinds/apm-python/pull/667/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L41) defined. Updates the tags for Lambda layer build too.

### DIFF
--- a/.github/workflows/build_publish_lambda_layer.yaml
+++ b/.github/workflows/build_publish_lambda_layer.yaml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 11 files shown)


cheempz (approved):
🎉

PR #681, opened 2025-08-13
NH-93748 Add ruff and other tox updates

tammy-baylis-swi:
Summary:

1. Adds `ruff` which is quicker and more comprehensive for code quality management than `pylint` -- we're still keeping the latter for now.
2. Removes a duplicate tox testenv, `py3{9,10,11,12,13}-lambda-gh`, which doesn't seem to be used anywhere.
3. Update other tox testenv to stop installing all deps with every single run. Only does it the first time now.

I've not applied Python code changes recommended by `ruff` here. That will be another PR. Same with adding `ruff` to ci/cd.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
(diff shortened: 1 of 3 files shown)


tammy-baylis-swi:
Will re-open

PR #688, opened 2025-09-04
NH-107164 Drop support of SW_APM_EXPORT_LOGS_ENABLED

tammy-baylis-swi:
Drops support of `SW_APM_EXPORT_LOGS_ENABLED` because it's a convenience for `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` which is already supported. This is a small breaking change for the next release; documentation changes in ticket.

Logs a special warning for the next release. In the future, it will be removed and the regular warning of "ignoring unsupported config" will be shown instead.


cheempz (approved):
LGTM but wondering what you think about making a breaking release (with the ready? API change) that also just drops support for this one entirely?

PR #689, opened 2025-09-04
NH-107699 Remove `integer_response` argument warning from SDK Ready function.

tammy-baylis-swi:
Removes the warning about `integer_response` argument if provided to APM SDK `solarwinds_ready` function.

Support was dropped at version 4.0.0 but we kept a courtesy warning to prevent argument errors. See ticket for documentation update.


cheempz (approved):
LGTM! One comment is we may want to release this as a breaking change.

PR #700, opened 2025-09-17
NH-119695 Potential fix for code scanning alert no. 451: Wrong name for an argument in a call

tammy-baylis-swi:
Potential fix for [https://github.com/solarwinds/apm-python/security/code-scanning/451](https://github.com/solarwinds/apm-python/security/code-scanning/451)

To fix the problem, we must ensure that only valid keyword arguments are passed when calling `get_apmconfig_mocks`. Since `md_is_valid` is not a parameter of `get_apmconfig_mocks`, we should replace the unsupported `md_is_valid=False` argument with an appropriate supported argument. The function parameters are: `mocker, enabled=True, is_lambda=False, export_metrics_enabled=True`. If the intention was to mock an invalid metric export configuration, then we should pass `export_metrics_enabled=False` instead. Only line 183 (and its surrounding context) within the fixture `mock_apmconfig_enabled_md_invalid` (lines 177-186) needs changing. No new imports or additional definitions are needed, as all required pieces exist.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/tests/unit/test_configurator/conftest.py
+++ b/tests/unit/test_configurator/conftest.py
@@ -180,7 +180,7 @@ def mock_apmconfig_enabled_md_invalid(mocker):
         "solarwinds_apm.configurator.SolarWindsApmConfig",
         get_apmconfig_mocks(
             mocker,
-            md_is_valid=False,
+            export_metrics_enabled=False,
         )
     )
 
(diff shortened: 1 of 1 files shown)


cheempz (dismissed):
LGTM but i think the entire `mock_apmconfig_enabled_md_invalid` can be removed, appears unused.

PR #759, opened 2026-04-16
Remove `boto` instrumentor dropped from OTel 1.41.0/0.62b0+

tammy-baylis-swi:
Removes k8s image pre-install and a test dependency on OTel Python `boto` instrumentor, which is no longer being released as of 1.41.0/0.62b0


cheempz (approved):
LGTM

PR #766, opened 2026-05-04
Update unit tests for traceflags 02,03

tammy-baylis-swi:
Updates unit test coverage as evidence that traceflags `02`, `03` are valid for existing APM Python custom sampling and propagation logic.

See ticket for more background information.


cheempz (approved):
LGTM

PR #768, opened 2026-05-15
Add Requests version constraint to image build

tammy-baylis-swi:
Adds a `requests` version constraint to image build, to work around Insecure Temporary File vulnerability and upstream instrumentor's too-open project.optional-dependencies.


cheempz (approved):
LGTM

PR #778, opened 2026-06-03
NH-135284 Upgrade APM to OTel Python 1.42.1/0.63b1

tammy-baylis-swi:
Upgrades APM to OTel Python 1.42.1/0.63b1. This:

* Drops Python 3.9 support
* Changes `otel.sdk.*` metrics into an opt-in feature with `OTEL_PYTHON_SDK_METRIC_ENABLED` (was always-on before)
* Can set traceflags as `00`, `01`, `02`, `03` (latter is "sampled + random")

Note we're skipping the minor version 1.42.0/0.63b0 bump.

What else is going on here:

The Amazon 2023 "Python 3.10" verify-install test was actually using py3.9 before. To maintain some coverage it's now an Amazon 2 py3.10 test with a dedicated workflow job "install-tests-amazon2" in order to get over glibc compatibility issues with Amazon 2.

Also drop CentOS 8 support entirely since EOL Dec 31 2021 -- it was an artifact for py3.9 coverage.

PR #780, opened 2026-06-04
Narrow version ranges for urllib3, requests in image publish

tammy-baylis-swi:
Narrows version ranges for urllib3, requests in image publish to address several CVEs with older versions.

Depends on drop of py3.9 support in https://github.com/solarwinds/apm-python/pull/778


cheempz (approved):
LGTM (enjoy your PTO copilot!)

PR #786, opened 2026-06-08
Update Lambda container example

tammy-baylis-swi:
Update the Lambda container example to show otelcol extension installation for x86_64 vs arm64.


cheempz (approved):
LGTM

PR #817, opened 2026-07-16
NH-143290 Update debug logging for Lambda

tammy-baylis-swi:
Updates APM Python debug logging to be less duplicative, and to be more clear/useful while in Lambda environment.

Changes:

* Cache and reuse Lambda-environment detection in `SolarWindsDistro` (`self._is_lambda`), and suppress specific debug logs when running in Lambda mode.
* Update `SolarWindsApmConfig.__str__()` to exclude service_key and collector in Lambda mode, while preserving masked service_key behavior outside Lambda; adds `is_lambda` to the output in both scenarios. This is logged at the end of ApmConfig init.
* Reduce repeated debug noise by logging “no default config file” only once; add/expand unit tests around these behaviors.

Also:

* Undo a change made by copilot in [3c8b6ec](https://github.com/solarwinds/apm-python/pull/817/commits/3c8b6ecaa9635d53263f51fc7b1c3ef434532604) which was making Lambda test publish fail

See ticket for additional testing.


solarwinds_apm/distro.py
```diff
@@ -171,9 +175,12 @@ def _configure(self, **kwargs: dict) -> None:
...
-                logger.debug(
-                    "Using exporter otlp collector endpoint %s", resolved
-                )
+                # See `wrapper` for Lambda settings
+                if not self._is_lambda:
+                    logger.debug(
+                        "Setting default OTEL_EXPORTER_OTLP_ENDPOINT %s",
+                        resolved,
+                    )
                 environ.setdefault(OTEL_EXPORTER_OTLP_ENDPOINT, resolved)
```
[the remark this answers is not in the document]
tammy-baylis-swi: This is minor, I want to leave it

solarwinds_apm/version.py
```diff
@@ -1 +1 @@
-__version__ = "7.0.0"
```
raphael-theriault-swi, line 1: Fourth digit means not a stable version if I understand correctly ? Just making sure it's not a typo but I think I remember seeing it a bunch in your release PRs

raphael-theriault-swi (approved):

jerrytfleung (approved):
LGTM

solarwinds_apm/version.py
tammy-baylis-swi: Yes that's correct 👍 

I actually didn't mean to commit this but testrelease is next anyway so it'll work.

PR #818, opened 2026-07-16
NH-143529 ResponseTime processor handles HTTP status code if non-compliant string

tammy-baylis-swi:
Adds ResponseTime metrics processor handling and debug logging if HTTP status code is a string, to prevent stacktrace from APM library.

String status code is non-compliant but can still happen easily in practice, e.g. AWS Lambda function responses. Status code _should_ be a 3-digit integer according to:

* [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-status-codes)
* [AWS WAF](https://docs.aws.amazon.com/waf/latest/developerguide/customizing-the-response-status-codes.html) (no Lambda equivalent doc but good links)
* [OTel semconv](https://github.com/open-telemetry/semantic-conventions/blob/99f629fec1a9e14a43a33ba8cc5ca7987e7c13f8/model/http/registry.yaml#L171-L175)


cheempz (approved):
LGTM

